PR #4, opened 2022-09-23
Add run list on model summary page

fivegrant:
Recreation of an older PR: https://github.com/jataware/phantom/pull/227


mattprintz (approved):
:+1:

PR #5, opened 2022-09-29
Include the `dojo-modeling/dojo` plugin system

fivegrant:
A copy + bug fix of the original plugin system.

Outstanding tasks:
- [ ]  Actually run and bug check the files
  - [ ] See if all the Causemos functionality works
- [ ] Implement more general 'transformation' action (e.g. dojo model schema to causemos model schema)
- [ ] Handle additional plugin data with pydantic schemas


mattprintz (approved):
:+1:

PR #6, opened 2022-09-30
Postgres api addition

marshHawk4:
Able to save/query datasets to/from a postgres database.


mattprintz:
@brandomr @marshHawk4 Is this still needed/wanted? I believe that we've decided not to go this route. Let me know if we want to keep/merge this or if I should just close it.

Sorrento110:
Closing this PR because we do not want to save datasets this way anymore.

PR #8, opened 2022-10-17
Adds Download CSV button to dataset summary page

marshHawk4:
This PR also added the wide/long download options to the Model Run summary page, but it turned out that model run downloads were broken - even the pre-wide/long version is no longer working. @fivegrant may have more context on why this is the case. For now, I just commented out the download button on the model run summary page so that we don't have a broken button.


mattprintz (approved):
Looks/works great

PR #11, opened 2022-10-18
Rename phantom/clouseau/claudine

mattprintz:
From `phantom/ui` to just `ui`.
From `clouseau`/`claudine`/`cato` to `terminal`/`connector`/`api`


adamgilg:
Phantom -> UI changes look good, and everything worked as expected when I ran it locally. Looks good to me!

PR #12, opened 2022-10-19
fix imports and allow configuration via args

brandomr:
@marshHawk4 @david-andrew please take a look at this. You can run it with something like:

```
python populate_wdi_data.py --es=http://localhost:9200 --aws_key='ACCESS_KEY' --aws_secret='SECRET'
```
except it doesn't work because `classDTO.StudentsDTO` on line 581 is not defined. Note that `pydantic` itself was not defined either and the `country.csv` file was missing so this would not have worked.

Also note that since we want this to be a utility, making things like the ES instance and AWS creds (as well as the bucket) configurable instead of hard coded is worth the extra 10 minutes of effort.

### DIFF
--- a/dojo/indicators/country.csv
+++ b/dojo/indicators/country.csv
@@ -1,0 +1,247 @@
+country_name,Alpha-2_Code,Alpha-3_Code,ISO_Numeric-3_Code
+Afghanistan,AF,AFG,4
+Åland Islands,AX,ALA,248
+Albania,AL,ALB,8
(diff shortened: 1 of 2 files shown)


mattprintz:
@brandomr Is this still needed?
I'm thinking we can close this since there hasn't been any movement in months and you state that it's not working.

Sorrento110:
Closing after discussion of current relevance.

PR #18, opened 2022-11-25
Datasets & Features Page + Semantic Search + Document Explorer

ccjoel:
Screen Shot 2022-11-29 at 10 26 45 AM


api/src/search/bert_search.py
```diff
@@ -0,0 +1,169 @@
...
+from .search import Search
+from src.datasearch.corpora import Corpus, T, Generic
+from typing import Union
+from transformers import BertTokenizer, BertModel, logging # type: ignore[import]
+from sentence_transformers import SentenceTransformer
+import torch
+from tqdm import tqdm
+
+
+class BertWordSearch(Search, Generic[T]):
```
david-andrew: You can get rid of this class, since you only use the sentence search class

tasks/tasks/search/bert_search.py
```diff
@@ -0,0 +1,170 @@
+# from __future__ import annotations
```
david-andrew: I see you have this file duplicated here, and at `api/search/bert_search.py`. Seems like you'd probably want to delete one, and pull everything from the other one

tasks/tasks/search/search.py
```diff
@@ -0,0 +1,11 @@
+# from __future__ import annotations
```
david-andrew, line 1: same note about this file being duplicated from `api/search/search.py`

PR #20, opened 2022-12-21
Fix published model version bump UI bug

adamgilg:
When trying to go through the model summary page version bump flow for a previously published model, there was a bug where the page would redirect/reload to the summary page URL for the old (published) model ID in the middle of the flow. This fixes that bug by preventing the form default behavior when the submit button ('Confirm') is clicked.

### DIFF
--- a/ui/client/components/SummaryIntroDialog.js
+++ b/ui/client/components/SummaryIntroDialog.js
@@ -88,7 +88,9 @@ const SummaryIntroDialog = ({
     setTimeout(() => setStep(defaultStep), theme.transitions.duration.standard);
   };
 
-  const versionBumpModel = async (relaunch) => {
+  const versionBumpModel = async (event, relaunch) => {
+    // prevent the page from redirecting on form submit
+    event.preventDefault();
     // set the endpoint for version bumping
     let versionBumpUrl = `/api/dojo/models/version/${model.id}`;
     let provisionUrl;
@@ -305,11 +307,11 @@ const SummaryIntroDialog = ({
                       disableElevation
                       className={classes.bigButton}
                       data-test="introDialogStartOverBtn"
-                      onClick={() => {
+                      onClick={(event) => {
                         if (model?.image) {
                           setStep('confirm');
                         } else {
-                          versionBumpModel();
+                          versionBumpModel(event);
                         }
                       }}
                     >
@@ -333,7 +335,7 @@ const SummaryIntroDialog = ({
                       color="primary"
                       disableElevation
                       className={classes.bigButton}
-                      onClick={() => versionBumpModel(true)}
+                      onClick={(event) => versionBumpModel(event, true)}
                       disabled={!model?.image}
                     >
                       <Grid container direction="row" justifyContent="center" alignItems="center">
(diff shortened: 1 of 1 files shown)


ui/client/components/SummaryIntroDialog.js
```diff
@@ -398,7 +400,7 @@ const SummaryIntroDialog = ({
                     type="submit"
                     disableElevation
                     disabled={disableConfirm}
-                    onClick={() => versionBumpModel()}
+                    onClick={(event) => versionBumpModel(event)}
```
ccjoel, line 403: No need to wrap the function with an anonymous arrow function.
Likely from copy-pasting or habit. Still happens to me occasionally.

ccjoel (approved):
Good find. I'll have a question for you that isn't relevant to merging this PR, I think.

ui/client/components/SummaryIntroDialog.js
```diff
@@ -398,7 +400,7 @@ const SummaryIntroDialog = ({
                     type="submit"
```
ccjoel, line 400: Quick question on the PR:

Is the reason that the default event page reload was occurring in the first place because the buttons were wrapped in a `<form>`? Is wrapping in a form, or using the `type=submit` prop necessary, then?

Basically an observation, that if not wrapping in form or not attaching a type=submit would also fix the problem, that might be a more direct fix than stopping the event propagation. At least to me, it is clearer that we don't wish to submit a form and can potentially prevent confusion/regressions.

Just a suggestion, but I think as it stands is good enough, and I might be missing something.

ui/client/components/SummaryIntroDialog.js
adamgilg: If the function isn't wrapped with an anonymous arrow function, it gets called when the JSX is calculated. For a button click we do need the wrapper or this happens the second the button shows up on the page. 
https://stackoverflow.com/questions/67529332/how-does-an-anonymous-function-work-in-onclick-in-react-component

ccjoel: Oh, no. I meant this:

```
onClick={versionBumpModel}
```

mattprintz (approved):
Good catch!

ui/client/components/SummaryIntroDialog.js
adamgilg: I added the `<form>` wrapper and `type=submit` a long while back because I thought it made for a better experience to be able to type the model name (the step you need to do to confirm you want to version bump) and then hit return/enter, instead of needing to click with your mouse. This probably matters more to me as the person who does it many times in a row developing the page than any real user, so it probably doesn't matter much either way. It isn't a bug fix that comes with any real downsides, however, so I think the marginal benefit in UX of being able to submit with enter is still worth it. 

Good question!

adamgilg: And it is a form, of sorts - just one that only has a single text field that we aren't saving any information from. We just confirm that their input matches the model name before enabling the submit button.

ui/client/components/SummaryIntroDialog.js
adamgilg: Ah, yes, that makes more sense!

PR #21, opened 2023-01-08
Two model summary page bugfixes

adamgilg:
This PR fixes two small model summary page bugs that I came across while working on Keycloak auth. 

1. The `?terminal=true` url param that we use to keep a user connected to the terminal UI (show them the back button, allow them to edit things on the summary page) was getting erroneously wiped clear because we were checking for the existence of the `lock` before it was getting loaded in. We're now waiting 10 seconds for the lock to load before clearing the terminal param and getting them off the 'running' summary page.
2. The model summary intro/version bump dialog was hiding any overflow due to a visual flicker MUI bug. Unfortunately this meant that on smaller screens (such as when developing with a console open) the buttons are totally hidden, with no way to click to the next dialog step. This is now fixed and the correct scroll is enabled, with the MUI bug still handled.

### DIFF
--- a/ui/client/summary.js
+++ b/ui/client/summary.js
@@ -54,19 +54,16 @@ const Summary = () => {
         });
     }
 
-    // save an empty variable so we can clear the timeout later on
-    let clearLock;
-    if (terminal === 'true' && !lock) {
-      // if we arrive with the terminal param but no lock
-      // set a timeout to get rid of it in 30 seconds (enough time to fetch a lock)
-      clearLock = setTimeout(() => history.replace(`/summary/${modelId}`), 30000);
-    }
+    const timeout = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


ui/client/summary.js
```diff
@@ -53,9 +53,19 @@ const Summary = () => {
           console.log('Container auto-shutdown sequence initiated');
         });
     }
-    // if we arrive with the terminal param but no lock, get rid of the terminal param
+
+    // save an empty variable so we can clear the timeout later on
+    let clearLock;
```
adamgilg: Try: checking inside the timeout if the conditions are met, then maybe we don't need to clear the timeout
Question: what's happening with the closure re clearlock? If we do it the current way, what about cleaning up the timeout?

mattprintz (approved):
Looks good

PR #22, opened 2023-01-09
Disable dataset column snapback

adamgilg:
After annotating a dataset column, the datagrid was automatically horizontally scrolling back to the start of the columns. This PR replaces the 'snapback' behavior with a success alert and leaves the user horizontally scrolled to the same position.


mattprintz (approved):
Looks good

PR #23, opened 2023-01-19
Mouseless Dataset table navigation

adamgilg:
This PR makes a number of small UI tweaks to the Dataset registration table view to allow for keyboard navigation. 
The changes are as follows:
- Click into any column to navigate between columns with arrow keys
- Get rid of single cell highlighting 
- Enter/return opens annotation panel
- Escape closes annotation panel
- Annotation panel opens with first input autofocused

There are also assorted visual tweaks along with this to make it all look better.


ccjoel (approved):

adamgilg:
Noticed during review: after annotating a far-right column, the focus stays on the column after it jumps to the start but the view stays far right. Perhaps shift focus to the column now in the place of the just-annotated column?

Sorrento110:
Merging this PR after discussion, the potential improvements mentioned above will be implemented later.

PR #33, opened 2023-02-24
Feature/search dataset features

ccjoel:
This is the up-to-date work for Document Explorer and Upload, as well as features index + Semantic Search for both documents and features.

I've removed the upload scripts folder for now, since after some cleanup they won't work as-is, and they were distracting on the PR. We can revisit if needed. They should be available on git history and the deployed analyst instance.

Once this lands into dev, I might re-add the upload scripts in some shape/form, as well as other improvements on a new, smaller, branch.


tasks/Dockerfile
```diff
@@ -19,7 +21,10 @@ RUN pip install numpy==1.22  # Numpy must be installed before GDAL (in requireme
 COPY ./tasks/requirements.txt /tasks/requirements.txt
 WORKDIR /tasks
 RUN pip install -r requirements.txt
-COPY ./tasks /tasks
 
+# TODO This did not exactly work:
```
ccjoel: Follow up on this. Maybe cache with shared volume as we did for causal recommender.

docker-compose.build-override.yaml
```diff
@@ -84,6 +84,7 @@ services:
     volumes:
       - ./api:/api
       - dataset-storage:/datasets
+      - ./documents:/documents
```
ccjoel, line 87: Is this the kind of thing that should be removed when envfile points to s3?

tasks/tasks/mixmasta_processors.py
```diff
@@ -319,7 +333,44 @@ def run_model_mixmasta(context, *args, **kwargs):
...
     shutil.rmtree(localpath)
 
-    response = {
-        "mixmaster_annotations": mm_ready_annotations
-    }
+    response = {"mixmaster_annotations": mm_ready_annotations}
     return response
+
+
+def scale_features(context, filename=None):
```
ccjoel: I should look into why this is here on my branch and not in dev.

tasks/tasks/mixmasta_processors.py
```diff
@@ -319,7 +333,44 @@ def run_model_mixmasta(context, *args, **kwargs):
...
+        file_buffer.seek(0)
+
+        put_rawfile(path=s3_filepath, fileobj=file_buffer)
+        data_paths_normalized.append(str(s3_filepath))
+    update_indicator(uuid=uuid, data_paths_normalized=data_paths_normalized)
+
+    return
+
+
+def update_indicator(uuid, data_paths_normalized):
```
ccjoel: This could be the only real change on this file. I think I was trying out a plugin and the quotes were all replaced to double quotes. I'll revert that the main change to be obvious here (PR).

ui/client/datasets/Flows.js
```diff
@@ -99,6 +99,20 @@ const BasicRegistrationFlow = {
...
   },
+  {
+    slug: 'scale',
+    title: 'Scaling Data',
+    label: 'Scaling',
+    component: RunJobs,
+    options: {
+      jobs: [
+        {
+          id: 'mixmasta_processors.scale_features',
```
ccjoel: Same for this flow entry- need to check how this made it into here.

ui/client/datasets/FileSelector.js
```diff
@@ -272,7 +272,7 @@ export const ExtraInput = ({
 /**
  *
  * */
-const FileInput = withStyles(({ spacing }) => ({
+export const FileInput = withStyles(({ spacing }) => ({
```
ccjoel, line 275: Ended up not using this, but no harm in exporting it.

ui/client/datasets/FormFields/index.js
```diff
@@ -33,10 +33,17 @@ export const FormAwareTextField = withStyles((theme) => ({
     }
   },
 }))(({
-  classes, name, label, requiredFn, placeholder, inputProps={}, InputProps={}, required, ...props
+  classes, name, label, requiredFn, placeholder, inputProps={}, InputProps={}, onBlur, required, ...props
```
ccjoel: Ended up not passing in an onBlur, but no harm for now. I might revert to simplify this component.

ui/client/datasets/RunJobs.js
```diff
@@ -87,6 +87,11 @@ const RunJobs = withStyles(({ spacing }) => ({
     }); // TODO catch , finally
   };
 
+  useEffect(() => {
```
ccjoel: Need to check how this made it here.

tasks/tasks/requirements.txt
```diff
@@ -1,5 +1,6 @@
 boto3==1.24.8
 elasticsearch==7.12.0
+fsspec==2022.11.0
```
ccjoel: Is this related to the feature scaling code I see in this diff? TODO follow-up

api/server.py
```diff
@@ -134,7 +136,7 @@ def setup_elasticsearch_indexes():
                         "type": "text",
                         "fields": {"keyword": {"type": "keyword", "ignore_above": 256}},
                     },
-                    "outputs": {"type": "object", "enabled": False},
+                    "outputs": {"type": "nested"},
```
ccjoel, line 139: Actually, now that we use the features index, this doesn't need to be nested anymore.

tasks/tasks/mixmasta_processors.py
ccjoel: Actually. Coming back fresh from the weekend, this file shouldn't have been modified at all. I'll issue a blame quickly to check how this change made it here, and revert anything related to feature scaling on this PR

ccjoel: Reverted PR that introduced unrelated changes.

mattprintz:
These changes were all merged in within PR #40 .

PR #36, opened 2023-02-27
Re-scale Indicator endpoint

marshHawk4:


api/src/indicators.py
```diff
@@ -507,3 +513,14 @@ async def create_preview(
             headers={"msg": f"Error: {e}"},
             content=f"Queue could not be deleted.",
         )
+
+
+@router.get("/indicators/{indicator_id}/rescale")
```
mattprintz: Actions like this should basically always be a PUT or POST. It's ok if there's no payload.
Since the publish endpoint is a PUT, this should probably be a PUT too.

api/src/indicators.py
```diff
@@ -507,3 +513,14 @@ async def create_preview(
...
+def get_all_indicator_info(indicator_id: str):
```
mattprintz: Function name is copy/pasted from above. Should be unique.

api/src/data.py
```diff
@@ -126,7 +126,7 @@ def cancel_job(job_id):
 
 # Last to not interfere with other routes
 @router.post("/job/{uuid}/{job_string}")
-def job(uuid: str, job_string: str, options: Optional[Dict[Any, Any]] = None):
+def runjob(uuid: str, job_string: str, options: Optional[Dict[Any, Any]] = None):
```
mattprintz: Not sure I love the name `runjob` because it both runs a job and checks/returns the status of running jobs.
I think `job` works for both cases, but I'm open to something else or an argument for `runjob`.

tasks/tasks/scale_indicator.py
```diff
@@ -0,0 +1,76 @@
...
+            feat = current_df[current_df["feature"] == f].copy()
+            current_min=np.min(feat["value"])
+            current_max=np.max(feat["value"])
+            if f not in min_max_mapping:
+                min_max_mapping[f]={"min":current_min, "max":current_max}
+            else:
+                if min_max_mapping[f]["min"] > current_min:
+                    min_max_mapping[f]["min"]=current_min
+                if min_max_mapping[f]["max"]<current_max:
+                    min_max_mapping[f]["max"]=current_max
```
mattprintz: Is this cleaner?
``` python
min_max_mapping[f] = {
    "min": min(
        current_min, 
        min_max_mapping.get(f,{}).get("min", current_min)
    ),
    "max": max(
        current_max,
        min_max_mapping.get(f,{}).get("max", current_max)
    ),
}

File: tasks/tasks/scale_indicator.py [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+            logging.info(f'downloading for s3 {path}')
+            download_rawfile(path, f"processing/{filename}")
+        except FileNotFoundError as e:
+            return {"success":False,'message':"File not found"}
+        
+        current_df=pd.read_parquet(f"processing/{filename}")
+
+        features = current_df.feature.unique()
+        for f in features:
+            feat = current_df[current_df["feature"] == f].copy()
```
mattprintz: copy is slow and potentially wastes a lot of memory. Can we not just work on `feat` and not modify it?

tasks/tasks/scale_indicator.py
```diff
@@ -0,0 +1,76 @@
...
+                min_max_mapping[f]={"min":current_min, "max":current_max}
+            else:
+                if min_max_mapping[f]["min"] > current_min:
+                    min_max_mapping[f]["min"]=current_min
+                if min_max_mapping[f]["max"]<current_max:
+                    min_max_mapping[f]["max"]=current_max
+    
+    logging.info(f'min max values for all files: {min_max_mapping}')
+
+    for path in data_paths:
```
mattprintz: We should check to see if min/max has changed and only overwrite the existing files if they have, otherwise we can simply save the new/missing normalized file(s).

tasks/tasks/scale_indicator.py
```diff
@@ -0,0 +1,76 @@
...
+        
+
+def scale_dataframe(dataframe, mapper):
+    """
+    This function accepts a dataframe in the canonical format
+    and min/max scales each feature to between 0 to 1
+    """
+    dfs = []
+    features = mapper.keys()
+    print(features)
```
mattprintz: Remove unnecessary print statements.

tasks/tasks/scale_indicator.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for f in features:
+        feat = dataframe[dataframe["feature"] == f].copy()
```
mattprintz: Another potentially slow copy. Try to find a way to avoid it.

mattprintz (changes requested):

tasks/tasks/scale_indicator.py
marshHawk4: Do we save min max values for features anywhere?
I am not aware that we will know which files are new additions or ones that have already been scaled with global min max. If we save current min max in elasticsearch I can see this working and its probably a good idea to store that, but I am not seeing those values.

mattprintz: New additions are files that exist in data_paths that do not (yet) have a corresponding _normalized entry.
We can calculate the min/max of the original files first, then the additions, and if the min max doesn't change then we still have to write the normalized files for the new additions, but don't have to do any work on the existing normalized files since there would be no change.

mattprintz:
@marshHawk4 Where are we updating the `data_paths_normalized` value in ES? Is that happening somewhere already?

tasks/tasks/scale_indicator.py
marshHawk4: Ok so currently my script normalizes the datasets and saves them in s3. I will just need to remove that code and modify the min_max mapping to compare the differences between the old verse new files.

Sorrento110:
Closing this PR in favor of PR #38

PR #37, opened 2023-02-27
0 to 1 feature scaling

Sorrento110:
PR to include 0 to 1 feature normalization step with Elwood.

### DIFF
--- a/tasks/tasks/mixmasta_processors.py
+++ b/tasks/tasks/mixmasta_processors.py
@@ -357,7 +357,7 @@ def scale_features(context):
             settings.DATASET_STORAGE_BASE_URL, "normalized", uuid, file_out_name
         )
 
-        file_buffer = io.StringIO()
+        file_buffer = io.BytesIO()
 
         dataframe_scaled.to_parquet(file_buffer)
         file_buffer.seek(0)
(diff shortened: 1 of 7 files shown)


tasks/tasks/mixmasta_processors.py
```diff
@@ -319,7 +334,56 @@ def run_model_mixmasta(context, *args, **kwargs):
...
+            "query": {"match": {"id": uuid}},
+            "fields": ["data_paths"],
+            "_source": False,
+        }
+        es_response = es.search(index="indicators", body=query)
+        data_paths = es_response["hits"]["hits"][0]["fields"]["data_paths"]
+    paths_to_process = [
+        path for path in data_paths if "_str" not in path and "_normed" not in path
+    ]
+    data_paths = [path for path in data_paths if "_normed" not in path]
```
mattprintz, line 357: I think this string should be `_normalized` instead of `_normed` based on the rest of the code, no?
But really, since we're storing the normalized files in a different field, I'm not sure we need this.

tasks/tasks/mixmasta_processors.py
```diff
@@ -319,7 +334,56 @@ def run_model_mixmasta(context, *args, **kwargs):
...
+
+    data_paths_normalized = []
+    for path in paths_to_process:
```
mattprintz, line 360: I don't think we can just process this one file at a time as we need to normalize to the global minimum/maximum not the local one in each file.

mattprintz (changes requested):
The `scale_features` task is very similar to the `scale_indicator` task in PR #36 . Please get with Kyle and determine the best way forward.

Sorrento110:
Closing this PR in favor of PR #38

PR #43, opened 2023-03-13
Overhaul to use Elwood during model execution

brandomr:
**Known issue**: the S3 push task in Airflow fails because Airflow isn't communicating with minio correctly. Otherwise this should work just fine.

**To do**: test that this didn't break anything in data registration flow.


tasks/Dockerfile
```diff
@@ -5,9 +5,9 @@ RUN apt-get update &&\
 
 RUN wget https://jataware-world-modelers.s3.amazonaws.com/gadm/gadm36_2.feather.zip
 RUN wget https://jataware-world-modelers.s3.amazonaws.com/gadm/gadm36_3.feather.zip
-RUN mkdir ~/mixmasta_data && \
-      unzip gadm36_2.feather.zip -d ~/mixmasta_data/ && \
-      unzip gadm36_3.feather.zip -d ~/mixmasta_data/ && \
+RUN mkdir ~/elwood_data && \
+      unzip gadm36_2.feather.zip -d ~/elwood_data/ && \
+      unzip gadm36_3.feather.zip -d ~/elwood_data/ && \
```
mattprintz, line 10: I think these lines need to stay as `mixmasta_data` until elwood is updated to look at that path:

https://github.com/jataware/elwood/blob/2afe2206b3d52e90d020f8749e0f3ffc6ae071a7/elwood/elwood.py#L267

mattprintz (changes requested):
Thank you for doing this. Looks like it should work, with the exception of the one exception I've highlighted here.

tasks/Dockerfile
brandomr: This is very interesting since it did work for me after building. I think maybe I didn't test `resolve to GADM` but I'm surprised it works otherwise.

brandomr: And bingo, checking `resolve to GADM` does cause failure. @Sorrento110 what do you think? We can either leave it as is in Elwood for now and change this `Dockerfile` back here or we can fix Elwood? I'd rather fix Elwood if it's not a big lift but open to ideas

Sorrento110: @brandomr I will change this in elwood, it's a really small lift.

Sorrento110: @mattprintz really nice catch!

mattprintz (approved):
Good to merge now.

PR #45, opened 2023-03-14
Semantic highlighter on Dojo API

ccjoel:


ui/client/components/SWRHooks.js
```diff
@@ -205,7 +205,11 @@ export function useDocuments(scrollId) {
   }
 
   const { data, error, mutate } = useSWR(
-    url, fetcher
+    url, fetcher, {
+      revalidateIfStale: false,
```
ccjoel, line 209: Bugfix: Clicking out and back to the browser window was forcing a refetch. We don't want this, as we may have performed a semantic search and may be displaying search results (we don't want those results to be replaced by a full documents refetch).

ui/client/documents/index.js
```diff
@@ -57,7 +59,7 @@ export const ConfidenceBar = withStyles((theme) => ({
 }))(LinearProgress);
 
 const semanticSearchParagraphs = async(query) => {
-  let url = `/api/dojo/paragraphs/search?query=${query}&size=50`;
+  let url = `/api/dojo/paragraphs/search?query=${query}&size=40`;
```
ccjoel, line 62: Since the page has good enough results count and both semantic search and semantic highlight may take "long" to respond, I have toned this down to 40 results. Especially while we try it the semantic highlighter on the deployed instance.

ui/client/documents/index.js
```diff
@@ -380,7 +355,17 @@ export const ParagraphTile = withStyles((theme) => ({
             <dt>
               <Chip classes={{root: classes.squareChip, label: classes.chipLabel}} label="text" />
             </dt>
-            <dd>{highlightText(paragraph.text, highlights)}</dd>
+            <dd>{(highlights || calculateHighlightTargets(paragraph.text, query))
```
ccjoel, line 358: Made it so that UI first highlights while awaiting for extended semantic highlighter results.

mattprintz (approved):
Looks great

PR #48, opened 2023-03-20
Env var disable semantic highlighting on UI

ccjoel:
Added DISABLE_SEMANTIC_HIGHLIGHT to .envfile (sample). The UI will disable API semantic highlight call, on Document explorer/search page, if it's value is `"true"`.

The UI fallbacks to keyword search calculated in the browser/js itself.

### DIFF
--- a/envfile.sample
+++ b/envfile.sample
@@ -40,3 +40,4 @@ UVICORN_RELOAD=True
 #DATASET_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/datasets/
 #DOCUMENT_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/documents/
 DOCUMENT_STORAGE_BASE_URL=file:///documents/
+DISABLE_SEMANTIC_HIGHLIGHT=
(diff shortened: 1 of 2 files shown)


envfile.sample
```diff
@@ -40,3 +40,4 @@ UVICORN_RELOAD=True
 #DATASET_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/datasets/
 #DOCUMENT_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/documents/
 DOCUMENT_STORAGE_BASE_URL=file:///documents/
+DISABLE_SEMANTIC_HIGHLIGHT=
```
brandomr: Can this default to `False` instead of empty?

brandomr (commented):
This looks ok to me. I'd be fine merging it if @mattprintz is OK with it and I can test it later this week before we actually deploy

ui/client/documents/index.js
```diff
@@ -452,21 +452,23 @@ const ViewDocumentsGrid = withStyles((theme) => ({
...
-
-          let url = `/api/dojo/paragraphs/highlight`;
-
-          return axios.post(url, {
-            query,
-            matches
-          })
-            .then((response) => {
-              setHighlights(response.data.highlights)
+          if (process.env.DISABLE_SEMANTIC_HIGHLIGHT !== "true") {
```
ccjoel, line 455: Only accepts "true" to disable, any other value will not disable semantic highlight, as requested.

envfile.sample
ccjoel: Yup- done! The value`=false` (or False, or anything else) will become a string, so the UI now compared to exactly to `"true"` (from =true in env file) in order to disable it (also suggested by Matt).

PR #52, opened 2023-03-31
Implemented highlight flag for paragraphs semantic search

ccjoel:
.. includes highlights in response upon request (off by default). Added response models for missing entries on Documents API.


mattprintz (approved):
Haven't tested, but code and logic all look good.

PR #53, opened 2023-04-03
Added endpoint to download original uploaded PDF file. Can implement …

ccjoel:
…displaying

original PDF using this endpoint.

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -31,6 +31,7 @@ requests==2.25.1
 python-dotenv==0.17.1
 PyYAML==5.4.1
 aioredis>=1.3.1,<1.4.0
+aiofiles==22.1.0
 fsspec==2022.8.2
 s3fs==2022.8.2
 torch==1.13.0
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -31,7 +31,6 @@ requests==2.25.1
 python-dotenv==0.17.1
 PyYAML==5.4.1
 aioredis>=1.3.1,<1.4.0
-aiofiles==22.1.0
 fsspec==2022.8.2
 s3fs==2022.8.2
 torch==1.13.0
--- a/api/src/documents.py
+++ b/api/src/documents.py
@@ -527,7 +527,9 @@ def upload_file(
 
 @router.get("/documents/{document_id}/file")
 def get_document_uploaded_file(document_id: str):
-
+    """
+    Downloads the original file (PDF) of an uploaded document.
+    """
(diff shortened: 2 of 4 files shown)


api/requirements.txt
```diff
@@ -31,6 +31,7 @@ requests==2.25.1
 python-dotenv==0.17.1
 PyYAML==5.4.1
 aioredis>=1.3.1,<1.4.0
+aiofiles==22.1.0
```
ccjoel: Library needed in order to return a `FileResponse`.

api/src/documents.py
```diff
@@ -522,3 +523,31 @@ def upload_file(
         },
         content=json.dumps({"id": document_id}|final_document),
     )
+
+
+@router.get("/documents/{document_id}/file")
```
ccjoel, line 528: So that users can retrieve the document on the S3 location.

api/src/documents.py
```diff
@@ -522,3 +523,31 @@ def upload_file(
...
+        logger.error(e)
+        return Response(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            headers={"msg": f"Error: {e}"},
+            content=json.dumps({"error": "Document has no uploaded source file."})
+        )
+
+    file = get_rawfile(path=s3_url)
+    headers = {'Content-Disposition': f'attachment; filename="{file_name}"'}
+    file_obj = file.detach()
```
ccjoel: Is there a better way to do this? I wasn't sure if I could return back a buffer as the response, so had to download the file from S3 to a tmp location, then return the obj location on response.

api/src/documents.py
```diff
@@ -522,3 +523,31 @@ def upload_file(
...
+    location = f"tmp/{file_name}"
+    with open(location, "wb") as new_file:
+        new_file.write(file_obj.read())
+
+    file_obj.close()
+
+    return FileResponse(location, media_type="application/octet-stream", filename=file_name)
```
mattprintz: Saving a temp file is probably not needed and could cause problems. Can't we just return a normal response with the headers and media type you already have there?

ccjoel: I wasn't able to, somehow. Do we do this anywhere else in the project? If not, Maybe we can quickly look at this together

ccjoel:
Merging this, as we went through and addressed feedback while in a huddle.

api/src/documents.py
ccjoel: Addressed.

PR #58, opened 2023-04-14
Features hybrid search

ccjoel:
Tried improving semantic search to make it display keyword matches first:
- Exact word matches (if search is single word) will show up first
- Exact sentence matches will follow (for more than one word search)
- Fuzzy search will be applied to search afterwards
- If none of the above work, semantic search will kick in

UI only performs the above hybrid search.

### DIFF
--- a/api/scripts/verify_missing_features.py
+++ b/api/scripts/verify_missing_features.py
@@ -26,19 +26,9 @@ es = Elasticsearch(f"http://{args.es_host}")
 print(es.info())
 
 
-total_processed = 0 # indicators
-# For outputs:
-total_missing = 0
-total_dups = 0
-
-
-def compareOutputsToFeatures(indicatorDictionary, indicator_id):
-    global total_missing
-    global total_dups
+def compareOutputsToFeatures(indicatorDictionary, indicator_id, totals):
 
     for index, output in enumerate(indicatorDictionary["outputs"]):
-
-        print(f"Processing output#{index}")
 
         feature_id = f"{indicator_id}-{output['name']}";
         body = {
@@ -51,34 +41,36 @@ def compareOutputsToFeatures(indicatorDictionary, indicator_id):
         total_found = res["hits"]["total"]["value"]
 
         if total_found == 0:
-            total_missing += 1
+            totals["missing"] += 1
         elif total_found > 1:
-            total_dups += 1
+            totals["duplicates"] += 1
+
+    return totals
 
 
-
-def parse_indicators(hits):
-    global total_processed
-
+def process_indicator_page(hits, totals):
     print(f"This page found {len(hits)} indicator hits.")
 
-    for idx, indicator in enumerate(hits):
+    for _, indicator in enumerate(hits):
         one_match = indicator["_source"]
         indicator_id = one_match["id"]
-        print(f"Indicators parsed: {total_processed}.\nProcessing indicator id: {indicator_id}\n")
-        compareOutputsToFeatures(one_match, indicator_id)
-        total_processed += 1
+
+        print(f"\nProcessing indicator with id: {indicator_id}\n")
+
+        compareOutputsToFeatures(one_match, indicator_id, totals)
+        totals["processed"] += 1
+
+    return totals
 
 
 SIZE = 10
 
-def process_indicators():
-    """
-    """
-    global total_missing
-    global total_dups
-
-    print("Called process_upload_indicators\n")
+def lookup_all_indicators():
+    totals = {
+        "processed": 0,
+        "missing": 0,
+        "duplicates": 0
+    }
 
     all_indicators = {
         "query": {
@@ -91,7 +83,7 @@ def process_indicators():
     hits = results["hits"]["hits"]
 
     if len(hits):
-        parse_indicators(hits)
+        totals = process_indicator_page(hits, totals)
 
     scroll_id = results["_scroll_id"]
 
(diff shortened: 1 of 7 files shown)


api/scripts/verify_missing_features.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python
+
+"""
```
ccjoel, line 3: Useful script to verify if there exist dataset.outputs that are not indexed in the features index.

api/validation/IndicatorSchema.py
```diff
@@ -510,6 +510,12 @@ class DateValidationResponseSchema(BaseModel):
         examples=[True, False],
     )
 
+class MetadataOpts(BaseModel):
+    match_score: float
```
ccjoel, line 514: Decided that any searches, either keyword or semantic, can and should include max_score (top-level results) and match_score (per result) properties

docker-compose.build-override.yaml
```diff
@@ -155,6 +155,8 @@ services:
     container_name: ui
     env_file:
       - envfile
+    environment:
+      NODE_OPTIONS: "--max-old-space-size=8192"
```
ccjoel, line 159: I hope this helps in making the webpack dev server not run out of heap memory when working with many (or big) files. Increased from ~4k.

ui/client/components/ViewFeatures.js
```diff
@@ -44,7 +45,7 @@ export const ConfidenceBar = withStyles((theme) => ({
 }))(LinearProgress);
 
 const semanticSearchFeatures = async(query) => {
-  let url = `/api/dojo/features/search?query=${query}`;
+  let url = `/api/dojo/features/search?query=${query}&size=50`;
```
ccjoel, line 48: I'm not sure why this was 10. On document search, I had bumped it some. Adding more results to allow for both:
a) Actual keyword/fuzzy sentence results first
b) Semantic results afterwards to fill up to 50 results

ui/client/components/ViewFeatures.js
```diff
@@ -219,56 +254,27 @@ const featureColumns = [
...
-
-  const [features, setFeatures] = useState([]);
-  const [featuresError, setFeaturesError] = useState(false);
-  const [featuresLoading, setFeaturesLoading] = useState(false);
 
   useEffect(() => {
     updateSearchTerm(searchTermValue);
   }, [searchTermValue]);
 
   const performSearch = () => {
```
ccjoel, line 262: Using hybrid search only now. No need to split keyword and semantic search.

ui/client/components/ViewFeatures.js
```diff
@@ -353,7 +335,7 @@ const ViewFeatures = withStyles((theme) => ({
         }}
         loading={featuresLoading}
         getRowId={(row) => `${row.owner_dataset.id}-${row.name}`}
-        columns={displayableColumns}
+        columns={featureColumns}
```
ccjoel, line 338: Displaying `Match %` column all the time, since it applies to any search (hybrid, the only one available now). Decided not to hide column when listing them.

api/scripts/verify_missing_features.py
Sorrento110: Is this script a one-off utility? I don't know if it is worth the refactor but it is generally good to avoid global variables and instead pass these count variables in from the parent functions and return them at the end of the child functions. If this script isn't that vital to the core process though it doesn't particularly warrant a refactor I don't think.

api/src/indicators.py
```diff
@@ -142,6 +140,71 @@ def patch_indicator(
     )
 
 
+def generate_keyword_query(term):
```
Sorrento110, line 158: I like ripping this out into it's own function to generate this big query, its more modular that way. Idk if it's worth it right now, but if we get another query this big we should move these generate_query type functions out into their own file, something like query_utils.py so that they don't clutter up the main indicators.py file.

api/src/indicators.py
```diff
@@ -277,13 +305,27 @@ def list_features(term: Optional[str]=None,
...
+    max_score = results["hits"]["max_score"]
+
+    def formatOneResult(r):
+        if term:
+            r["_source"]["metadata"]={}
+            r["_source"]["metadata"]["match_score"] = r["_score"]
+        r["_source"]["id"] = r["_id"]
+        return r["_source"]
+
+    response = {
```
Sorrento110, line 329: The two response variables here (the return object and ES response) would benefit from having distinct names to improve readability. Maybe es_response and features_payload?

Sorrento110 (commented):
All of these changes look good to me, my comments are minor improvements for readability or best practices.

api/scripts/verify_missing_features.py
ccjoel: It's a one-file utility script. Right now it is a dev utility, but has the potential to be run in production if we want to occasionally sync datasets.outputs with the features index- the first step would be to check if the two are out of sync!

I don't foresee us using this anytime soon, but I'll give that refactor a try 👍🏼

ccjoel: Addressed.

api/src/indicators.py
ccjoel: Addressed- es results are now `es_hits`.

api/src/indicators.py
ccjoel: We'll address at a later time. Good suggestion though!

PR #66, opened 2023-05-02
Fixes Document Upload file list/PDF view mismatch

adamgilg:
There was a bug where the selected file in the left hand files list on /documents/upload wouldn't match the displayed PDF. We were matching the files to their blob url with an array index, which turned out to not be reliable. This switches to matching them with the file path. 

Currently there's nothing safeguarding against multiple files with the same path, or capping the number of files. Both of these can be added to this PR or done in a future PR.


ccjoel (approved):
+1

PR #67, opened 2023-05-03
Adds 10 file limit for document upload

adamgilg:
This PR adds a 10 file limit for the document uploader. It's possible to circumvent the limit by adding up to 9 files, then adding up to 10 more. However, with the 10 file cap per drag&drop/selector, and the entire uploader disabled after it passes 10, the maximum a user can reach is 19 with this workaround. 

I also added alert-based error messaging when the file picker/drag & drop fails, either because too many documents were added or for another reason (screenshot below). 

Screenshot 2023-05-03 at 6 36 25 PM

### DIFF
--- a/ui/client/documents/upload/DropArea.js
+++ b/ui/client/documents/upload/DropArea.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { useDropzone } from 'react-dropzone';
 import clsx from 'clsx';
 import { withStyles } from '@material-ui/core/styles';
@@ -6,6 +6,8 @@ import { withStyles } from '@material-ui/core/styles';
 import IconButton from '@material-ui/core/IconButton';
 import InboxIcon from '@material-ui/icons/MoveToInbox';
 import Typography from '@material-ui/core/Typography';
+
+import BasicAlert from '../../components/BasicAlert';
 
 const extensionMap = {
   'pdf':{
@@ -31,9 +33,6 @@ export function formatExtensionForDropZone(extensionsArray) {
   }, {});
 }
 
-/**
- *
- **/
 export const FileDropSelector = withStyles((theme => ({
   root: {
     margin: '1rem 0 1rem 0',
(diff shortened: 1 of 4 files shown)


ui/client/documents/upload/DropArea.js
```diff
@@ -62,15 +61,77 @@ export const FileDropSelector = withStyles((theme => ({
...
+  acceptExtensions,
+  mini,
+  CTA,
+  multiple = true,
+  disableSelector,
+}) => {
+  const [alertVisible, setAlertVisible] = useState(false);
+  const [alertMessage, setAlertMessage] = useState('');
+
+  const onDropFilesRejected = (fileRejections) => {
```
ccjoel, line 76: I have two observations about this code- the first impression is that this component is reused elsewhere in the app, which is not related to documents ( :-0 ). We have to check both the effects it causes on the other usage, as well as the error/warning texts associated with documents and are not generic as it stands.

I'll comment on the second one a bit lower on this diff.

ccjoel: (I know the component is defined under `documents/`. We may move up to common `component/` folders now that it is reused; that is my bad)

ui/client/documents/upload/DropArea.js
```diff
@@ -62,15 +61,77 @@ export const FileDropSelector = withStyles((theme => ({
...
+      setAlertMessage('The upload limit is 10 files. Please try again with fewer files.');
+      setAlertVisible(true);
+    } else {
+      setAlertMessage(fileRejections[0].errors[0].message);
+      setAlertVisible(true);
+    }
+  };
 
-  const {getRootProps, getInputProps, isDragActive} = useDropzone({
+  const { getRootProps, getInputProps, isDragActive } = useDropzone({
```
ccjoel, line 92: Second observation (mentioned earlier on a comment)- `useDropzone` hook  seems to accept `disabled`, `noClick`, and/or `noDrag` properties. Is it possible to use that in this instance?

https://react-dropzone.js.org/#src

ccjoel: There is also a `maxFiles`. Did you try this?

ccjoel (changes requested):
I left some comments of aspects I think we can improve. I'm mostly available through slack (besides standup) if you'd like to chat about them.

ui/client/documents/upload/DropArea.js
adamgilg: Ah, I searched for its usage elsewhere but I must have accidentally only searched within `/documents/upload`. I'll explore how this interacts with the other use.

ui/client/documents/upload/DropArea.js
adamgilg: I used disabled and maxfiles (both of which are a few lines below this comment), but I didn't see the noDrag property. I'll give that a try and see if it disables the drag and drop functionality. Without the eventlistener, even when disabled, if you drag a file onto the input it will then open the pdf as the page itself. Definitely not the behavior we want. I'll check out noDrag.

adamgilg: On further exploration, it looks like `disabled` is indeed disabling the react-dropzone drag and drop behavior already, so noClick/noDrag aren't necessary. However, the eventListeners are necessary to prevent the default browser drag and drop behavior (which works anywhere on the site, but why would someone be dragging a PDF onto the landing page, I guess?), because we still have the input area that _looks_ like it should accept drag and drop. In my testing I didn't even realize it was the default browser behavior for a while, because I just kept thinking dropzone wasn't fully disabled and was instead misbehaving in some way. If we're disabling upload, I think it's pretty critical to do this step as well.

ui/client/documents/upload/DropArea.js
adamgilg: I moved the error messaging for the Dataset use of DropArea into the same alert, as seen below. It still has the capability to display an error inline (ie not in the BasicAlert) if it gets one when processing, via [this line](https://github.com/jataware/dojo-stack/blob/1b8f5af0b1f95bb9ca67bf94cb40f3a6a301085a/ui/client/datasets/annotations/Table/index.js#L304). It looks and functions the same in all other ways here. Without the `disabled` prop being passed in, nothing else will change.

Screenshot 2023-05-05 at 4 09 36 PM
Screenshot 2023-05-05 at 4 09 28 PM

ui/client/documents/upload/DropArea.js
ccjoel: Oooh I see what you mean. Users will run into this awkward situation of dropping the file on the page (not the drop zone) and navigating away from Dojo into, possibly, a PDF preview page! Good catch!

ui/client/documents/upload/DropArea.js
ccjoel: Awesome, thanks! That works a lot better now.

ccjoel:
Looks good- I appreciate that you addressed both usages. Merging.

ccjoel (approved):

PR #68, opened 2023-05-09
Delete files in Document Uploader

adamgilg:
This PR adds delete buttons to the files in the document uploader. The functionality is straightforward, with a single caveat: it will currently always reset the selected file to the first file in the list. Because of the way we're managing the files (two different arrays for the files and the metadata) and the selected file (a reference to the index), this seemed like the simplest and safest way of handling the variety of cases where a user may delete the selected file, delete a file before the selected file (thus changing where the index is pointing), etc. We can do a more complex handling of this if we want, but I think anything else will be more likely to introduce bugs. 

Considering that there's a 10 file limit (and that any edits to the metadata form persist when switching between files), this shouldn't be taxing for users.

### DIFF
--- a/ui/client/documents/upload/FileList.js
+++ b/ui/client/documents/upload/FileList.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import { withStyles } from '@material-ui/core/styles';
 
-import { formatBytes } from '../utils';
-
+import ClearIcon from '@material-ui/icons/Clear';
+import IconButton from '@material-ui/core/IconButton';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemIcon from '@material-ui/core/ListItemIcon';
@@ -13,10 +13,9 @@ import Paper from '@material-ui/core/Paper';
 import RadioGroup from '@material-ui/core/RadioGroup';
 import Radio from '@material-ui/core/Radio';
 
-/**
- *
- **/
-export const FileTile = withStyles((theme) => ({
+import { formatBytes } from '../utils';
+
+export const FileTile = withStyles(() => ({
   root: {
     cursor: 'pointer'
   },
@@ -36,7 +35,13 @@ export const FileTile = withStyles((theme) => ({
   filenameText: {
     wordBreak: 'break-all'
   }
-}))(({ classes, file, value, uploadStatus, onClick, selected, onDelete }) => {
+}))(({
+  classes, file, value, onClick, selected, onDelete
+}) => {
+  const handleDeleteClick = (event) => {
+    event.stopPropagation();
+    onDelete();
+  };
 
   return (
     <ListItem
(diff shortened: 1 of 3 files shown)


ui/client/documents/upload/FileList.js
```diff
@@ -13,10 +13,9 @@ import Paper from '@material-ui/core/Paper';
 import RadioGroup from '@material-ui/core/RadioGroup';
 import Radio from '@material-ui/core/Radio';
 
-/**
- *
```
ccjoel, line 17: Hey- is there a particular reason you're removing these? I know the jsdoc-like comments are empty right now, but I'd rather add some description here than remove the jsdoc placeholder.

To share my reasoning: It both adds a description, but I also take advantage that it already separates symbol/function definitions in a way that gets me more room between each other while going through the file:





but of course that's just a cherry on top- the end-goal is to eventually for me (or anyone else that runs into this!) to fill in or add whatever's useful for them (if there isn't anything already).

No big deal. Curious to hear your thoughts on my approach to fn padding and docstrings and/or if it bothers you that they're sometimes empty.

ui/client/documents/upload/index.js
```diff
@@ -249,21 +246,16 @@ const UploadDocumentForm = withStyles((theme) => ({
...
-   * TODO: In order to handle deletes- either do backflips around deleted indexes,
-   * not rendering deleted ones, etc, or change the collection/representation
-   * of data from arrays and selected index to an object with key=>properties.
-   **/
   const handleFileDelete = (index) => {
-    // TODO Implement me once we use SortedMap
-    console.log("deleting file index:", index);
+    // always reset back to zero, even if we aren't touching the selectedFile
+    // so we don't end up in a weird state where we're modifying the array around this index
+    setSelectedFileIndex(0);
```
ccjoel, line 251: I was worried about the difference between 0 and null here. When there's one file remaining and we remove it, the index goes back to 0 (instead of null) and we're pointing to no selected file. At one point, during the initial implementation of document upload,  there weren't checks and this would blow up. But I see there are checks below- so this works!

Maybe `selectedFileIndex` should default to `0` instead of `null` (when initializing that portion of state), to make it cleaner now (and remove any null checks for it), but that's not relevant to this PR so we can address it whenever we'd like.

ccjoel (approved):
I'm so glad that you figured and implemented this one. Sometimes, somehow, I get implementation-exhaustion and don't immediately address these important details on the initial implementation (also because it was never an explicit requirement, but it is an obvious and important UX aspect nonetheless).

So- thank you, thank you.

ccjoel:
I went through the PR but will let you merge it, in case you'd like to take any actions on the very-optional notes I left here.
Thanks again for making the Document Uploader a better place ;)

ui/client/documents/upload/FileList.js
adamgilg: I use an editor that runs eslint and shows me the errors, so I generally clean up small things that throw linter errors when I'm in a file. Empty comments are one such thing. I don't feel strongly about them either way, though I'd both lean more towards separating things out into separate files to make them more legible and just adding a comment when there is actually one to be added, rather than preemptively putting space in for it. 

(and separating things out into different files is a totally subjective, not at all all-the-time thing, so even my feelings on that are complicated)

But like I said, I don't feel strongly about it. I do like having a linter and style guidelines, as I find it helps me focus and not need to make little decisions about minor inconsequential things. Plus the linter does catch actual issues (such as missing dependencies in hooks). But that's a whole separate discussion 😄. 

All of which is to say, I'm happy to stop deleting those from the top of functions/components. I'll look into adding an exception to the linter that will make it stop complaining about those specific empty comments, rather than all empty comments... And feel free to let me know if any of my other linting-inspired cleanup when I'm in a file changes something you would rather not change!

ui/client/documents/upload/index.js
adamgilg: Ah good call, I'll test that out and see if I can simplify things. Thanks for the suggestion!

adamgilg:
> I went through the PR but will let you merge it, in case you'd like to take any actions on the very-optional notes I left here. Thanks again for making the Document Uploader a better place ;)

Don't merge just yet. I'm going to spend a few minutes playing around with your idea for initializing selectedFileIndex to 0 and removing the null checks.

ui/client/documents/upload/FileList.js
ccjoel: Cool, thanks- if it isn't a lot of effort it'd be nice to either not remove them, or move to their own file if you think that'd be better. I'd been assessing some of these files while going back- some of them are a bit long by now (but haven't done much about it yet).

adamgilg: Adding them back! And Eslint was just throwing a fit because of a lack of spacing after the comment started, so I added an exception for that.

adamgilg:
Good call with the `selectedFileIndex` @ccjoel, it feels much cleaner to default to zero and remove the null checks. I also added back the comments and put in an edit to the eslint rules so it won't complain about the empty jsdoc style comment blocks.

PR #72, opened 2023-05-12
All geo annotations should be set to resolve_to_gadm=true and checkbox UI hidden

ccjoel:
I still have an unknown in my mind, which is:

Should absolutely all geo types get the resolve_to_gadm set to true, even for geo types where we didn't allow that to be set (checkbox was not shown or disabled)? Same for fiends that have the explicit required "geocoding level" dropdown field for users to change- it is harmless to add resolve_to_gadm=true to those under the cover?

Example UI:
Screenshot from 2023-05-11 16-37-07

I tried one test dataset registration and it worked (api nor elwood broke), although I hope there wasn't a semantic bug in there after this change.

The hard way to verify this is going through all the unit tests diff for [dataOUT.test.js](https://github.com/jataware/dojo-stack/pull/72/files#r1192517357) and verify that the outputs make sense, given the geo type/categories (lat/lon/coordinates/country/territory/municipality, etc).


ui/client/datasets/annotations/dataOUT.test.js
```diff
@@ -20,7 +20,7 @@ describe('formatAnnotationsOUT', () => {
         geo_type: 'latitude',
         coord_format: 'lonlat',
         primary: false,
-        resolve_to_gadm: false,
+        resolve_to_gadm: true,
```
ccjoel, line 23: dataOUT.test.js checks for resolve_to_gadm=true for all geo fields, to validate that we'll always send that property to the backend.

ui/client/datasets/annotations/ColumnAnnotation.js
```diff
@@ -20,6 +20,14 @@ import { ExternalLink } from '../../components/Links';
 import { removeSelf } from './annotationRules';
 import { GEO_ADMINS } from './constants';
 
+
+// <FormAwareCheckBox
```
ccjoel, line 24: I can remove once we've decided to finalize the changes here.

ui/client/datasets/annotations/ColumnAnnotation.js
```diff
@@ -72,8 +80,8 @@ export const GeoControls = ({
   const isCoorPairCandidate = ['latitude', 'longitude']
     .includes(values.geo_type);
   // country, state/territory, county/district:
-  const isResolveGADMCandidate = [GEO_ADMINS.admin0, GEO_ADMINS.admin1, GEO_ADMINS.admin2]
-    .includes(values.geo_type);
+  // const isResolveGADMCandidate = [GEO_ADMINS.admin0, GEO_ADMINS.admin1, GEO_ADMINS.admin2]
```
ccjoel, line 83: Same, can clean up after getting other eyes on the feature request and its effects.

brandomr:
@ccjoel resolve GADM is only relevant with categorical places (e.g. country, admin1-3) and not relevant for lat/lons. We shouldn't be providing the resolve to GADM flag for lat/lon or if we do it's just ignored by elwood. @Sorrento110 could confirm that.

ccjoel:
I think it'll be easier/faster to close this one, and add a PR that only does this on UI:

> resolve GADM is only relevant with categorical places (e.g. country, admin1-3)

Will do so after handling the urgent items for today.

PR #74, opened 2023-05-15
Added causemos plugin example to envfile.sample.

ccjoel:
Pending items:

- Add mock `test-app` docker service to received plugin/published messages. Possibly rename for something more meaningful for this purpose. Covered by open #89 

## Questions

- I'm not completely familiar with the causemos publish setup- does `DEBUG=true` cause it to skip publishing altogether?
Addressed on a call and the answer is YES- debug=true would historically cause causemos to not be notified (which we want for local dev)
- Should we add an entry to dojo-stack/README or api/README on how to setup and/or use the publishing plugin?
The answer is YES- we'll addres either on #89 or after it is merged.

### DIFF
--- a/envfile.sample
+++ b/envfile.sample
@@ -41,3 +41,4 @@ UVICORN_RELOAD=True
 #DOCUMENT_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/documents/
 DOCUMENT_STORAGE_BASE_URL=file:///documents/
 DISABLE_SEMANTIC_HIGHLIGHT=false
+PLUGINS={"causemos":"src.plugins.causemos.CausemosPlugin","logger":"src.plugins.logging.LoggingPlugin"}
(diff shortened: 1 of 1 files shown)


ccjoel:
Leaving this here so that we don't forget

```
"sync":"src.plugins.sync.SyncPlugin"
```

Sorrento110:
Added the Sync plugin to this!

envfile.sample
```diff
@@ -40,4 +40,5 @@ UVICORN_RELOAD=True
 #DATASET_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/datasets/
 #DOCUMENT_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/documents/
 DOCUMENT_STORAGE_BASE_URL=file:///documents/
-DISABLE_SEMANTIC_HIGHLIGHT=false
\ No newline at end of file
+DISABLE_SEMANTIC_HIGHLIGHT=false
+PLUGINS={"causemos":"src.plugins.causemos.CausemosPlugin","logger":"src.plugins.logging.LoggingPlugin", "sync":"src.plugins.sync.SyncPlugin"}
```
ccjoel, line 44: These are the values set in the deployed .envfile, but it's never bad to have them as examples here.

PR #75, opened 2023-05-15
Fixes to the indicators not having outputs during dataset registration.

Sorrento110:
This is a fix to repair the process of taking each feature in the annotations and appending them to the indicator outputs.

### DIFF
--- a/ui/client/datasets/Flows.js
+++ b/ui/client/datasets/Flows.js
@@ -92,6 +92,7 @@ const BasicRegistrationFlow = {
       jobs: [
         {
           id: 'elwood_processors.run_elwood',
+          send_context: true,
           handler: async ({result, annotations, setAnnotations, datasetInfo, setDatasetInfo, ...extra}) => {
             const updatedDataset = {
               ...datasetInfo,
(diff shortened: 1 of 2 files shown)


ui/client/datasets/Flows.js
```diff
@@ -92,6 +92,7 @@ const BasicRegistrationFlow = {
       jobs: [
         {
           id: 'elwood_processors.run_elwood',
+          send_context: true,
```
ccjoel, line 95: I took a peek in the original changes where we started using elwood instead of mixmasta, but this line was still missing before then. I'm missing some context. @Sorrento110 it would be helpful for me to understand why this wasn't needed before (did "mixmasta" cache the context previously?)

Let me know! Thanks.

Sorrento110: I think it has more to do with the change to elwood processors where there is now an else clause at the bottom of the code chunk that appends features when it can't find a qualfies object in the features object in the annotations: https://github.com/jataware/dojo-stack/blob/815a335cd37d1f4d419e3d93112e3532687219b8/tasks/tasks/elwood_processors.py#L259

ccjoel: Aaah, I missed that completely. Thanks!

ccjoel: I take it this send_context is not needed, but harmless, then.

ccjoel:
#76 include these changes. Closing (thanks!)

PR #76, opened 2023-05-15
Added new sync plugin

Sorrento110:
This syncs the annotations and indicators on publish. This also fixes the issue where the indicator outputs are not updated from the annotations and gives us a job we can run on prior uuids to restore their outputs.

### DIFF
--- a/api/src/plugins/sync.py
+++ b/api/src/plugins/sync.py
@@ -1,0 +1,124 @@
+import json
+
+import requests
+from fastapi.logger import logger
+from src.plugins import PluginInterface
+from src.indicators import update_indicator
+from src.data import get_context
+from validation import IndicatorSchema
+
+
+class SyncPlugin(PluginInterface):
+
+    # Sync annotations to indicators
+    def publish(self, data, type="indicator"):
+        if type == "indicator":
+            uuid = data["id"]
+            context = get_context(uuid)
+
+            # Outputs
+            qualifier_outputs = []
+            outputs = []
+            feature_names = []
+            for feature in context["annotations"]["annotations"]["feature"]:
+
+                feature_names.append(
+                    feature["name"]
+                )  # Used for the primary qualifier outputs.
+                output = dict(
+                    name=feature["name"],
+                    display_name=feature["display_name"],
+                    description=feature["description"],
+                    type=feature["feature_type"],
+                    unit=feature["units"],
+                    unit_description=feature["units_description"],
+                    ontologies={},
+                    is_primary=True,
+                    data_resolution={
+                        "temporal_resolution": context.get("dataset", {}).get(
+                            "temporal_resolution", None
+                        ),
+                        "spatial_resolution": context.get("dataset", {}).get(
+                            "spatial_resolution", None
+                        ),
+                    },
+                    alias=feature["aliases"],
+                )
+                # Append
+                # TODO: Hackish way to determine that the feature is not a qualifier
+                if feature.get("qualifies", None):
+                    if len(feature["qualifies"]) == 0:
+                        outputs.append(output)
+                        # Qualifier output for qualifying features
+                    elif len(feature["qualifies"]) > 0:
+                        qualifier_output = dict(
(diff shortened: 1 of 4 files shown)


api/src/settings.py
```diff
@@ -42,7 +42,8 @@ class Settings(BaseSettings):
     PLUGINS: Dict[str, str] = {
         # "my_plugin": "plugin_module.MyPlugin",  # Where plugin_module.MyPlugin is an importable dotted path and MyPlugin
         # is a subclass of utils.PluginInterface
-        # "causemos": "src.plugins.causemos.CausemosPlugin",
+        "causemos": "src.plugins.causemos.CausemosPlugin",
+        "sync": "src.plugins.sync.SyncPlugin",
```
ccjoel, line 46: I think this gets overwritten by prod's plugin entry (which is missing this new `sync` property). We should do the same suggestion as before- add the plugins+=`sync` entry by default on envfile.sample.

I'm thinking basically what this PR did:

https://github.com/jataware/dojo-stack/pull/74/files#diff-146894aeacf42e1a854d258950e085970851940b8967ebb32cb2fe16c1875393R44

But maybe we go ahead add add both updates as part of this PR (since it is easy to forget to add this on the envfile both on local and prod).

Let me know if it doesn't make sense otherwise 👍🏼

api/src/plugins/sync.py
```diff
@@ -0,0 +1,124 @@
...
+                    is_primary=True,
+                    data_resolution={
+                        "temporal_resolution": context.get("dataset", {}).get(
+                            "temporal_resolution", None
+                        ),
+                        "spatial_resolution": context.get("dataset", {}).get(
+                            "spatial_resolution", None
+                        ),
+                    },
+                    alias=feature["aliases"],
```
ccjoel, line 45: As aliases are somewhat optional- would it make sense to change this to a `dict.get()` call? I'm worried about users forgetting to add this property when using csv file annotation or other mechanisms.

ccjoel (commented):
I left some questions while I test this out

ccjoel:
We should confirm that model output doesn't have the same issues (or address them), I thjink. Unless it doesn't apply, which would be nice.

ccjoel:
I merged since it's started to get to eod and this is time-sensitive. @Sorrento110 please see my questions above. If they make sense I can apply the changes myself, I just wanted a second opinion before doing so.

tasks/tasks/elwood_processors.py
```diff
@@ -231,7 +231,7 @@ def run_elwood(context, filename=None, on_success_endpoint=None):
         )
         # Append
         # TODO: Hackish way to determine that the feature is not a qualifier
-        if feature["qualifies"]:
+        if feature.get("qualifies", None):
```
ccjoel, line 234: We should review why this was occurring, outside of csv dictionary annotations.

ui/client/datasets/Flows.js
```diff
@@ -92,6 +92,7 @@ const BasicRegistrationFlow = {
       jobs: [
         {
           id: 'elwood_processors.run_elwood',
+          send_context: true,
```
ccjoel, line 95: We should also review and understand what this does (I, myself, not 100% sure).

tasks/tasks/elwood_processors.py
ccjoel: This can be refactored:
- One if/ one else
- Invert condition: if it qualifies, add the qualifier logic output, else add outputs
- Ensure test condition works as expected for both None and empty array []

ccjoel: And we know the answer to the original question- empty list `[]` is falsey, so the whole section was being skipped instead.

ui/client/datasets/Flows.js
ccjoel: We discussed on a call. Useful for the model outputs annotation flow. Not needed for dataset registration, although it won't do any harm unless there is a race condition involved.

PR #82, opened 2023-05-17
Document Explorer Column Sorting

adamgilg:
With the new server-mode pagination, the document explorer sorting was no longer working. This PR gets it working again with changes to the UI and API code. 

The Document Explorer <DataGrid> component now has custom handling for when the sort changes, which hooks into the same data fetching logic as the pagination. We now, however, pass along sort_by and order parameters to `GET /documents`.

The `GET /documents` endpoint now receives these optional `sort_by` and `order` parameters, checks if they are one of the valid fields, and if so, will return the sorted list of documents as requested. 

I also broke `documents/index.js` into several files and did some general cleanup for code organization/readability, as it was starting to get fairly long and confusing to work in.

### DIFF
--- a/api/src/documents.py
+++ b/api/src/documents.py
@@ -15,6 +15,7 @@ from fastapi import (
     UploadFile,
     File,
     Request,
+    Query,
 )
 from fastapi.responses import FileResponse
 from fastapi.logger import logger
@@ -282,9 +283,24 @@ def format_document(doc):
 @router.get(
     "/documents/latest", response_model=DocumentSchema.DocumentListResponse
 )
-def latest_documents(scroll_id: Optional[str]=None, size: int = 10):
+def latest_documents(
+    scroll_id: Optional[str] = None,
+    size: int = 10,
+    sort_by: str = Query("uploaded_at", description="Field to sort by"),
+    order: str = Query("desc", description="Order to sort by")
+):
     """
     """
+    if order not in ["asc", "desc"]:
+        raise HTTPException(status_code=400, detail="Invalid sort order. Must be either 'asc' or 'desc'")
+
+    valid_fields = ["title", "publisher", "description", "uploaded_at", "creation_date"]
+    if sort_by not in valid_fields:
+        raise HTTPException(status_code=400, detail=f"Invalid sort field. Must be one of {valid_fields}")
+
+    # Apply .keyword to specific fields
+    if sort_by in ["title", "publisher", "description"]:
+        sort_by += ".keyword"
 
     q = {
         "query": {
@@ -292,8 +308,8 @@ def latest_documents(scroll_id: Optional[str]=None, size: int = 10):
         },
         "sort": [
             {
-                "uploaded_at": {
-                    "order": "desc"
+                sort_by: {
+                    "order": order
                 }
             }
         ]
@@ -310,7 +326,6 @@ def latest_documents(scroll_id: Optional[str]=None, size: int = 10):
         scroll_id = None
     else:
         scroll_id = results.get("_scroll_id", None)
-
     return {
         "hits": results["hits"]["total"]["value"],
         "items_in_page": totalDocsInPage,
--- a/api/src/documents.py
+++ b/api/src/documents.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from typing import Any, Dict, Generator, List, Optional
 from urllib.parse import urlparse
 from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import RequestError
 from fastapi import (
     APIRouter,
     HTTPException,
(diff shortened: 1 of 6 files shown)


ui/client/documents/ParagraphTile.js
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
```
ccjoel, line 1: YES YES YES

api/src/documents.py
```diff
@@ -323,17 +324,38 @@ def latest_documents(scroll_id: Optional[str]=None, size: int = 10):
...
+    size: int = 10,
+    sort_by: Optional[str] = Query(None, description="Field to sort by"),
+    order: Optional[str] = Query(None, description="Order to sort by")
+):
     """
     Retrieves all stored documents, regardless of files uploaded or processed.
+    Optional `sort_by` and `order` params return sorted results.
     """
 
+    if order not in ["asc", "desc"]:
```
ccjoel, line 339: This will be a problem when using the API directly (which Uncharted's causemos app does).
sort_by and order are optional, but then order is required on this line. We should provide defaults for `sort_by` and `order`, and only validate it if `sort_by` was provided.

api/src/documents.py
```diff
@@ -323,17 +324,38 @@ def latest_documents(scroll_id: Optional[str]=None, size: int = 10):
...
+        raise HTTPException(status_code=400, detail="Invalid sort order. Must be either 'asc' or 'desc'")
+
+    valid_fields = ["title", "publisher", "description", "creation_date"]
```
ccjoel: On the API side, we should probably allow sorting by any fields (even if the UI will only sort by these). Unless there are issues with arbitrary `sort_by`. If so, let's talk about those. I can help!

(Uncharted's app will likely want to sort by something different)

ui/client/documents/index.js
```diff
@@ -639,6 +385,10 @@ const ViewDocumentsGrid = withStyles(() => ({
                 onPageChange={handlePageChange}
                 paginationMode="server"
                 rowCount={totalRowsCount}
+                disableColumnFilter
+                sortingMode="server"
+                onSortModelChange={handleSortChange}
```
ccjoel, line 390: I noticed that the datagrid does not display a sort arrow when it loads (but it is sorted by `creation_date` by default). It then adds the sort arrow when explicitly selecting fields to sort (clicking on their sort arrow). We should probably display the sort arrow by default, which shouldn't harm and should inform users of the default sort order.

## Currently
#### When it first loads:

Screenshot from 2023-05-18 10-50-38


#### After manually selecting sort order:

Screenshot from 2023-05-18 10-50-43

Would this cause unexpected side effects? What do you think?

ui/client/documents/index.js
```diff
@@ -639,6 +385,10 @@ const ViewDocumentsGrid = withStyles(() => ({
...
+                disableColumnMenu
```
ccjoel, line 391: Nit/nice to have (brainstorming out loud):

I wonder if we should add the column menu, if this is what I think this is. That is, it would be nice to users to select which columns to display if they can, and allow to sort by any of those. Or maybe this isn't what I think it is.

Also, this is very optional for our purposes, but in general if we were doing any other UI and this was easy to do it would be very nice (especially as the API would allow sorting by arbitrary fields- see my previous comments).

ccjoel (changes requested):
First off- thanks for splitting the documents/index file!

I left some comments. Some pointing out some problems, others a bit more optional.

I actually could help with some of these changes- I thought of just modifying them without leaving review comments, but this is a nice "todo" list to track as well as I didn't want to modify and push commits if you might be working on some changes in parallel.

Let me know, and we could even split up some of these. Thanks!

(I also realize this is a relatively more "picky" review than our usual ones. No specific reason at all I suppose, except that I felt inspired to improve the ui sorting and I'm offering to help! I suppose that counts.)

ui/client/documents/index.js
adamgilg: Yeah I think this is a good idea.

Edit: although, a further thought: this might make it confusing about what column the grid is currently being sorted by. In order to do this we'll need to have a custom header anyways, so it might need to be an greyed out arrow or something like that and then the sorted one is black. I'll play around with it and see what works.

ui/client/documents/index.js
adamgilg: Removing the column menu was actually the very last thing I did, for a few reasons:
1. There are only four columns at the moment, so I didn't think we need to allow the removing columns thing. If Brandon wants to include more columns in the future, this might be worth reconsidering.
2. Filtering isn't enabled on the backend, so we don't have filtering actually working (another default feature of the column menu).
3. Most significantly, if you have the sort buttons (ASC/DESC in the menu) enabled at all, you have to include the "unsort" button. This doesn't do anything at the moment and I don't really even know what unsorted would mean. Since we already have the arrow, including the menu would basically just be giving us redundant ASC/DESC buttons and a nonfunctional unsort button. 

It was something I thought about and tried to fix for a while though! In the end the broken unsort button sealed the fate of the menu for me - I'd rather not have a menu with a broken button and questionable utility when the main purpose of it is in that arrow.

ccjoel: Ah, I see. Nevermind then. I just realized another cool idea related to this (which we could do fro all our grids if we wanted to), but I'll share it in slack later.

api/src/documents.py
adamgilg: Yeah good catch - I was just testing this with no `sort_by` or `order` and it currently fails.

PR #83, opened 2023-05-18
UI Dataset Register: Clear `time_format` for `date_type=epoch`

ccjoel:
cartwright was adding an invalid value there (inferred value broke registration), which isnt used but somehow was confusing elwood. Fixed and added ui unit test to confirm it.

This was what UI wsa loading as default value:

Screenshot from 2023-05-18 10-29-26

Then I proceeded to switch the subcategory to `epoch` (instead of `Date`), but forgot to clear `date format`

when clicking next and running the annotation processing through elwood, we would get:

Screenshot from 2023-05-18 10-39-45

Hopefully this isn't a bug anymore after this PR 🎉

### DIFF
--- a/ui/client/datasets/annotations/dataOUT.js
+++ b/ui/client/datasets/annotations/dataOUT.js
@@ -125,6 +125,13 @@ function formatDateAnnotationOUT(localAnnotation, outgoingAnnotationBase) {
     genDateMultiPartMemberAnnotation(localAnnotation, outgoingAnnotation)
       .forEach((item) => { collectedOutgoingAnnotations.push(item); });
   }
+
+  if (localAnnotation.date_type == 'epoch') {
+    // fix for cartwrights/geotimeclassify inserting time_format for epochs, which is not needed
+    // also helps with csv dictionary annotation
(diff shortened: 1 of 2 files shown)


ui/client/datasets/annotations/dataOUT.js
```diff
@@ -125,6 +125,13 @@ function formatDateAnnotationOUT(localAnnotation, outgoingAnnotationBase) {
     genDateMultiPartMemberAnnotation(localAnnotation, outgoingAnnotation)
       .forEach((item) => { collectedOutgoingAnnotations.push(item); });
   }
+
+  if (localAnnotation.date_type == 'epoch') {
+    // fix for cartwrights/geotimeclassify inserting time_format for epochs, which is not needed
+    // also helps with csv dictionary annotation
```
ccjoel, line 131: (meaning, helps in the case someone enters invalid, unnecessary, time_format(s) on their csv,xls dictionary annotation file column).

Sorrento110:
So the inferred value was 'unix_time'? This change looks simple to fix this issue!

Sorrento110 (approved):

PR #84, opened 2023-05-23
Small changes to Document Explorer table

adamgilg:
This updates the following in the Document Explorer table/grid: 

- Removes the Description column from the table and adds description to each document's dialog 
- Adds uploaded_at as a column
- Renames uploaded_at and creation_date to `Date Created` and `Date Uploaded`

### DIFF
--- a/ui/client/documents/ViewDocumentDialog.js
+++ b/ui/client/documents/ViewDocumentDialog.js
@@ -61,7 +61,7 @@ export const ViewDocumentDialog = ({ doc, onClose }) => {
         {!isEmpty(doc) && (
           <dl>
             {['publisher', 'creation_date', 'type', 'original_language',
-              'classification', 'producer', 'stated_genre', 'id']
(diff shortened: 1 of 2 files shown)


mattprintz (approved):

ccjoel (changes requested):
I ran this and realized that @adamgilg doesn't have access to proper uploaded documents (macbooks won't really run the text extraction/embedding within a docker container). Looks like we need to format the values for `Date Uploaded` when displaying them.

Take a look:

Screenshot from 2023-05-24 10-32-03

The rest is fantastically outstanding.

ps. Maybe I need to share a backup of my es docs//paragraphs index with @adamgilg. We can talk more about something like that later.

adamgilg:
Ahh good catch. I'll ping you tomorrow about this @ccjoel

adamgilg:
Fixed it with the same date formatting as Date Created.

Screenshot 2023-05-30 at 4 57 39 PM

ccjoel (approved):
Awesome!

PR #87, opened 2023-05-24
Store data transformations on annotations object

adamgilg:
This adds the data transformation selections to the annotations object when the user hits `NEXT` on the Transformation step. The data is nested under `transformation` and is identical to the arguments that are sent to kick off the Elwood jobs. Each transformation is keyed under the Elwood job name: `clip_geo`, `regrid_geo`, etc. 

I also added `transformations` to the MetadataSchema under `annotations.transformations`. This can easily be moved if anyone thinks it would make more sense under metadata or under its own top level field.

### DIFF
--- a/api/validation/MetadataSchema.py
+++ b/api/validation/MetadataSchema.py
@@ -174,6 +174,12 @@ class ClipTime(BaseModel):
 
 Transformation = Union[RegridGeo, ScaleTime, ClipGeo, ClipTime]
 
+class Metadata(BaseModel):
+    class Config:
+        extra = Extra.allow
+
+    transformations: Optional[Dict[str, Transformation]]
+
 class AnnotationSchema(BaseModel):
     class Config:
         extra = Extra.allow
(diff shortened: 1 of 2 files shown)


Sorrento110:
This PR is working for me, storing the transformations. I think storing them under the "annotations" key makes sense but if we wanted to keep that key clean for just the dataset annotations we could move them to the "metadata" key or we could make another new top-level key for transformations. There's not really a benefit or detriment one way or the other as far as I can see.

Sorrento110 (approved):

api/validation/MetadataSchema.py
```diff
@@ -150,6 +181,7 @@ class Config:
     geo: Optional[List[GeoAnnotation]]
     date: Optional[List[DateAnnotation]]
     feature: Optional[List[FeatureAnnotation]]
+    transformations: Optional[Dict[str, Transformation]]
```
Sorrento110: Something in this is causing the transformation annotations to not store anymore. I am taking a look, but I think the format we are feeding the annotations in do not match this validator format.

Sorrento110 (changes requested):

api/validation/MetadataSchema.py
adamgilg: This all works locally for me. I'm going to mark this as resolved for now but happy to revisit if it continues to be an issue.

PR #88, opened 2023-05-25
Fixes to the dataset registration to add current date into the dataset

Sorrento110:
Registration flow changes for datasets with no date annotation. Enables registration of datasets with no date.

### DIFF
--- a/api/src/indicators.py
+++ b/api/src/indicators.py
@@ -479,15 +479,16 @@ def patch_annotation(payload: MetadataSchema.MetaModel, indicator_id: str):
         body = json.loads(payload.json(exclude_unset=True))
 
         # Handles datasets being regsitered with no date.
-        if not body.get("annotations").get("date", []):
-            logger.info("No Date Annotated, making one.")
-            rawfile_path = os.path.join(
-                settings.DATASET_STORAGE_BASE_URL,
-                indicator_id,
-                f"raw_data.csv",
-            )
-            date_annotation = add_date_to_dataset(path=rawfile_path)
-            body["annotations"]["date"] = [date_annotation]
+        if body.get("annotations"):
+            if not body.get("annotations").get("date", []):
+                logger.info("No Date Annotated, making one.")
+                rawfile_path = os.path.join(
+                    settings.DATASET_STORAGE_BASE_URL,
+                    indicator_id,
+                    f"raw_data.csv",
+                )
+                date_annotation = add_date_to_dataset(path=rawfile_path)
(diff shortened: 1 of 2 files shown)


api/src/indicators.py
```diff
@@ -486,10 +478,21 @@ def patch_annotation(payload: MetadataSchema.MetaModel, indicator_id: str):
 
         body = json.loads(payload.json(exclude_unset=True))
 
+        # Handles datasets being regsitered with no date.
+        if not body.get("annotations").get("date", []):
```
ccjoel: This fails when a PATCH request does not contain an `annotations` entry. For example, when submitting file metadata (files, filename, selected sheet, etc) when first registering a dataset.

api/src/indicators.py
```diff
@@ -486,10 +478,21 @@ def patch_annotation(payload: MetadataSchema.MetaModel, indicator_id: str):
...
+                indicator_id,
+                f"raw_data.csv",
+            )
+            date_annotation = add_date_to_dataset(path=rawfile_path)
+            body["annotations"]["date"] = [date_annotation]
+
         es.update(index="annotations", body={"doc": body}, id=indicator_id)
 
         return Response(
-            status_code=status.HTTP_201_CREATED,
```
ccjoel, line 492: Good catch!

ccjoel (changes requested):
Good stuff, but saw a potential bug when `annotations` isn't present on the patch annotations request. Left a comment there- body["annotations"] in that scenario doesn't exist.

api/src/indicators.py
Sorrento110: yea I'll split this out to check if "annotations" exists. The second get will fail here since the first will default to None and calling .get() on a None type won't work.

ccjoel (commented):
Worked for me while testing locally 👍🏼

api/src/indicators.py
```diff
@@ -486,10 +478,22 @@ def patch_annotation(payload: MetadataSchema.MetaModel, indicator_id: str):
 
         body = json.loads(payload.json(exclude_unset=True))
 
+        # Handles datasets being regsitered with no date.
+        if body.get("annotations"):
+            if not body.get("annotations").get("date", []):
```
ccjoel, line 481: Actually, I just realized: the UI warns the user that a missing `primary` date will default to today's date, but what should we do if the user did annotate a date field, but is not marked a primary? I don't think this code handles that scenario:

Screenshot 2023-05-29 at 7 05 55 PM

ccjoel (approved):

PR #90, opened 2023-05-29
Fix UI integration tests

ccjoel:
Fixed all dataset and model runs tests. Have not addressed Models tests as I haven't worked on those and they seem to be more involved.


ccjoel:
Will wait to hear feedback on the View Model Run page section disappearing (on chat) before merging.

ccjoel:
An integration test for Model Run Summary caught something missing in the UI. I'll revert the one change relevant as we decide what we'll do for the missing section (data paths or download csv button for a model run).

PR #91, opened 2023-05-29
[DRAFT]: Pass one at updating developer docs.

ccjoel:
Very likely that we'll need many more improvements, potentially author new readme/contributing/and dedicated task docs from scratch.


api/src/settings.py
```diff
@@ -25,11 +25,11 @@ class Settings(BaseSettings):
     DOCKERHUB_PWD: str = ""
     DOCKERHUB_ORG: str = "jataware"
 
-    DATASET_STORAGE_BASE_URL: str = "file:///datasets/"
+    DATASET_STORAGE_BASE_URL: str = "s3://datasets/"
```
ccjoel, line 28: @mattprintz do you know a reason why this would be a bad idea? `file:///` was failing for us today, and we figure we can get rid of using that altogether to work around the problems we ran into.

envfile.sample
```diff
@@ -39,6 +38,8 @@ UAZ_HITS=10
 UVICORN_RELOAD=True
 #DATASET_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/datasets/
 #DOCUMENT_STORAGE_BASE_URL=s3://jataware-world-modelers-dev/documents/
-DOCUMENT_STORAGE_BASE_URL=file:///documents/
+CONFIG_STORAGE_BASE=s3://dojo/configs/
```
ccjoel, line 41: @mattprintz same for these values- are these Ok defaults for dev/prod? `file:///` was giving me issues today.

ccjoel: Also `CONFIG_STORAGE_BASE` was missing from envfile.sample.

@Sorrento110 what do you think of these changes?

ccjoel:
These changes have worked reliably on local dev, and settings.py/envfile.sample values need to be modified anyways for prod/stage deployments. Merging.

PR #95, opened 2023-06-02
Prevents the model summary page from reloading the page in the middle…

ccjoel:
… of the new model version flow when trying to version bump a published model

Re-created this one to be based off dev, since it was previously only merged to main.


adamgilg (approved):

adamgilg:
Awesome, thanks for doing this!

PR #105, opened 2023-06-06
Fixes for Model Publishing

ccjoel:
To test locally, ensure these env vars are set:

- CAUSEMOS_IND_URL=http://testmos:8012/api/maas
- CAUSEMOS_DEBUG=false

### DIFF
--- a/api/src/plugins/causemos.py
+++ b/api/src/plugins/causemos.py
@@ -9,58 +9,57 @@ from src.ontologies import get_ontologies
 from src.plugins import PluginInterface
 from validation import ModelSchema
 
+from src.settings import settings
+from src.models import model_versions
+
 
 class CausemosPlugin(PluginInterface):
     def publish(self, data, type):
+        if settings.DEBUG:
+            return
+
         if type == "model":
-
-            from src.settings import settings
-
-            from .models import model_versions
-
-            if settings.DEBUG:
-                return
-
-            notify_data = deepcopy(data)
-            notify_data = convert_to_causemos_format(notify_data)
+            notify_data = convert_to_causemos_format(data)
 
             # On the notification step only, we want to include any previous versions so that they can be deprecated
             previous_versions = model_versions(data["id"])["prev_versions"]
+
+            logger.info(f">>> previous_versions: {previous_versions}")
+
             notify_data["deprecatesIDs"] = previous_versions
 
+            logger.info(f">>> notify_data: {notify_data}")
+
             # Notify causemos of the new model
-            self.notify_causemos(data=notify_data, type="model")
+            self.notify_causemos(data=notify_data, entity_type="model")
 
             # Send CauseMos a default run
-            logger.info("Submitting defualt run to CauseMos")
+            logger.info("Submitting default run to CauseMos")
             self.submit_run(notify_data)
+
         if type == "indicator":
-            from src.settings import settings
-
-            if settings.DEBUG:
-                return
-
             # Notify causemos of the new indicator
             # logger.info("causemos indicator", json.dumps(data, indent=2))
-            self.notify_causemos(data=data, type="indicator")
+            self.notify_causemos(data=data, entity_type="indicator")
 
-    def notify_causemos(self, data, type="indicator"):
+
+    def notify_causemos(self, data, entity_type="indicator"):
         """
         A function to notify Causemos that a new indicator or model has been created.
 
-        If type is "indicator":
+        If entity_type is "indicator":
             POST https://causemos.uncharted.software/api/maas/indicators/post-process
             // Request body: indicator metadata
 
-        If type is "model":
+        If entity_type is "model":
             POST https://causemos.uncharted.software/api/maas/datacubes
             // Request body: model metadata
         """
         headers = {"accept": "application/json", "Content-Type": "application/json"}
 
-        if type == "indicator":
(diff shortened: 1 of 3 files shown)


api/src/plugins/causemos.py
```diff
@@ -9,58 +9,57 @@
...
+
 
 class CausemosPlugin(PluginInterface):
     def publish(self, data, type):
-        if type == "model":
-
-            from src.settings import settings
-
-            from .models import model_versions
+        if settings.DEBUG:
```
ccjoel, line 18: Moved this to here, as its repeated logic between datasets and models.

api/src/plugins/causemos.py
```diff
@@ -9,58 +9,57 @@
 from src.plugins import PluginInterface
 from validation import ModelSchema
 
+from src.settings import settings
```
ccjoel, line 12: Importing on top of file instead of on the `publish` definition. I don't know why it was the other way before- hope I'm not missing anything.

api/src/plugins/causemos.py
```diff
@@ -9,58 +9,57 @@
...
 
+        if type == "indicator":
             # Notify causemos of the new indicator
             # logger.info("causemos indicator", json.dumps(data, indent=2))
-            self.notify_causemos(data=data, type="indicator")
+            self.notify_causemos(data=data, entity_type="indicator")
 
-    def notify_causemos(self, data, type="indicator"):
+
+    def notify_causemos(self, data, entity_type="indicator"):
```
ccjoel, line 42: changed to entity_type here to not overshadow the python `type` method. Did not replace the other `type` occurrences as they seemed more involved.

api/src/plugins/causemos.py
```diff
@@ -156,13 +157,7 @@ def post_update(self, data, type):
...
 
 
-def convert_to_causemos_format(self, model):
-    """
-    Transforms model from internal representation to the representation
-    accepted by Cauesmos.
-    """
-
-    def to_parameter(annot):
+def to_parameter(annot):
```
ccjoel, line 156: Moved definition out of the other method- this can be declared once.

api/src/plugins/causemos.py
```diff
@@ -184,6 +179,12 @@ def to_parameter(annot):
             "max": float(annot["max"]) if annot["max"] != "" else None,
         }
 
+
+def convert_to_causemos_format(model):
```
ccjoel, line 179: This had a reference to `self` that would not work, since it isn't part of a class and was expecting 2 params, receiving one when called.

api/src/plugins/sync.py
```diff
@@ -118,17 +118,17 @@ def publish(self, data, type="indicator"):
...
-            "id": uuid,
-            "name": context["dataset"]["name"],
-            "description": context["dataset"]["description"],
-            "maintainer": context["dataset"]["maintainer"],
-            "outputs": outputs,
-            "qualifier_outputs": qualifier_outputs,
-            "feature_names": feature_names,
-            "published": True,
-        }
+            response = {
```
ccjoel, line 121: This was outside the check for datasets being synced, which is not what we want, I think. It was being run on model publish and failing, do to variables for patch_indicator not being defined per the check for dataset type before on line 15.

ccjoel:
Setting multiple reviewers as I'm unfamiliar with this are of the code, as well as to spread the wisdom on how publishing/sync is working.

@jgawrilo you may even review after it has been merged if you'd like (or not- optional).

 I think I might add a pytest test for the transform before and after the model publish step (looks like we massage the model from what we have stored to what we send to causemos). No need to block these changes as they are a bit of a hotfix (model publishing not working on either deployed instance).

api/src/plugins/causemos.py
```diff
@@ -9,58 +9,57 @@
...
-
-            notify_data = deepcopy(data)
-            notify_data = convert_to_causemos_format(notify_data)
+        if type == "model":
+            notify_data = convert_to_causemos_format(data)
 
             # On the notification step only, we want to include any previous versions so that they can be deprecated
             previous_versions = model_versions(data["id"])["prev_versions"]
+
+            logger.info(f">>> previous_versions: {previous_versions}")
```
ccjoel: !! Need to remove these logs before merging

api/src/plugins/causemos.py
mattprintz: Because the plugins get instantiated as part of the settings (we can decide which plugins to run based on env variable which turns in to a setting which turns in to instances of the plugins).
Having the import be inside the classes was intended to prevend circular imports.

mattprintz (commented):
Looks good, assuming we're confident we're not creating circular imports.

api/src/plugins/causemos.py
ccjoel: Not merging until we are done with the tests and address this.

mattprintz (approved):
Approved. Looks good. Thanks

api/src/plugins/causemos.py
ccjoel: Discussed on slack, and this concern is addressed by:

https://github.com/jataware/dojo/blob/main/api/src/plugins/__init__.py#L31-L43

which supersedes the previous method of importing within the class's method to circumvent circular dependencies.

That is, plugins already load lazily as part of the change shared above.

PR #106, opened 2023-06-07
Lint fixes

adamgilg:
Bundling all the outstanding lint fixes into one PR for easy testing.


ui/client/datasets/annotations/MultiColumnGeoSelector.js
```diff
@@ -23,12 +23,16 @@ const MultiColumnGeoSelector = ({
   // we hardcode the column's value (adminX option) to the columName being edited
   // and even disable that field (see disabled prop on the react tree below).
   useEffect(() => {
+    // TODO: I can't figure out how to hit even the console.log outside of the if statement
```
adamgilg: @ccjoel I can't figure out how to hit this useEffect - maybe you have an idea?

ui/client/datasets/AnomalyDetection.js
```diff
@@ -65,19 +65,19 @@ const RunJob = withStyles(({ spacing }) => ({
 }) => {
```
adamgilg, line 65: Is this component used anymore? It doesn't seem to be imported anywhere

ui/client/datasets/Annotate.js
```diff
@@ -99,7 +99,7 @@ export default withStyles(({ spacing }) => ({
   };
 
   useEffect(() => {
-    if (!isLoading || (isLoading && !datasetInfo?.id)) {
+    if (!isLoading || (isLoading && !datasetInfo?.id) || isEmpty(annotations?.metadata)) {
```
adamgilg, line 102: This works in my manual testing, but there are so many edge cases with the annotation page. I'm assuming we're always loading in metadata, but if this isn't the case then we'll need another way of solving this. Without this check the rest of this effect gets run twice, as the annotations object first loads empty, then gets populated. Which means we hit the various endpoints in this useEffect twice (not the end of the world, but better not to).

ui/client/datasets/annotations/MultiColumnGeoSelector.js
ccjoel: This is not for multi column using two coordinates (lat lon)

but instead for a multipart annotation that consists of 4 columns, one of each:
country, admin1, admin2, and admin3

If you have a dataset with 4 geo columns that match that, and you start an annotation in one where you select one of them to be of any of those types (not coordinates nor latitude or longitude) you can toggle on the checkbox to start this process and render this component.

ccjoel: image

ui/client/datasets/AnomalyDetection.js
ccjoel: Not used anymore- this was a PoC for an additional step on the dataset registration process, which would detect anomalies, antipatterns in uploaded datset files (eg too many empty rows, some others) and inform the user.

I don't remember if we explicitly or implicitly completely scratched it out. I recall a decision to temporarily disable. I'm not sure if we're interested in revisiting at some point.

cc @mattprintz , @brandomr

ui/client/datasets/Annotate.js
ccjoel: Hm. I'm not sure. As I recall `annotations.metadata` can load empty or populating when getting to this page. And it can be for either:
- Model Output Annotation
- Dataset Register
- Dataset Update

Among others. I'll checkout the branch and try this out so that I can jumpstart my memory and run various tests scenarios.

ccjoel:
I can see that the cypress integration tests already failed from these changes- we'll have to look into it.

ccjoel:
Woah! UI Lint job passed! @adamgilg deserves some heavy bitcoin!

PR #107, opened 2023-06-07
Load Large-Language Models Just-in-time

ccjoel:
This is the work I've mentioned previously that would allow us to easily build, start, and dev reload the API during local development. It should also improve the build and startup speed of our production stack (as the PR stands now- more on this soon).

This is particularly important for individuals running the stack and developing on a mac laptop with our docker setup. Especially especially useful if working on the Dojo API (as the dev reload time on any file change is greatly improved).

## What we did:

1) Instead of a singleton that gets initialized as soon as the API starts, we export a getter (eg `get_embedder`) that returns a singleton instance and initializes it if hasn't initialized before. That is, the plain-and-simple singleton pattern.

2) Instead of caching the downloaded models within the api docker image, we set the environment variable that indicates Torch and Sentence_transformers where to store their cache. Then, we set a shared docker volume between host and container, which means that completely deleting and rebuilding the dojo api image and starting the fresh container will already have the model data cached (shared with its host!). Meaning, the model is not downloaded again, but step 1) above is related to loading the model into memory, which still incurs some time cost (when first requesting a semantic search endpoint after every server reload, if not forced to hot-load).

## Questions:

1) I noticed an `api/docker-compose-dev.yaml` file. Should we move the shared volume setup there and only use this caching strategy to help us during development, but not cache the LLM models in this volume on deployments/prod?

2) As it stands, the API will start quickly, but the first semantic search/highlight will load the model in place and that adds some overhead to the query time (per the just-in-time changes), but only on the first semantic search per api startup (not often in prod).

Would we like to only use this to help us on dev-only, or maybe add a prod-helper-script (or code to server.py) that issues a short request to these semantic search endpoints as soon as the API is up and ready to hot-load the required models instead? This causes the LLM models to load AFTER the api starts, instead of during, which I presume is non-blocking (the API could start responding to requests while it also hotloads the models).

## How I tested this

I tried all these changes both on Ubuntu and macos docker locally.

[1] I completely removed old api images, rebuilt, restarted the api and used semantic search. Then I went ahead and rebuilt and restarted he api once more, checked that it started quickly, then ran a semantic search and checked the logs and time it took. The logs only show the api loading the model data into memory, but not downloading it! (which is what we want).

I stopped the services, removed the api image, and restarted the services once more (which builds the api container image from scratch). I saw the same results as when I performed item [1] above.

Note: My linux desktop won't sweat at all regardless, but on the macos even loading the models (without downloading them!) incurs some cost (they load slowly the first time, but not as terribly as before per the shared-volume caching now).

=======

# Miscellaneous changes

You'll notice this PR has several poetry-python related lockfiles that make it look huge. I included some cleanup for helper scripts under `api/scripts`, as well as the addition of the script that populates the document_paragraph index with dart sample data. This will allow anyone to use the Document Explorer and perform semantic search without having to upload their own documents (which is extremely slow and failure-prone when running on a local macos-docker setup).

The reason I included it as part of this work is that I had to populate my DB indeces from scratch again in order to test the api semantic search endpoints.

*Feel free to ignore anything related to the `api/scripts` folder if it is distracting, as that's never used in production regardless.*

If anyone has opinions on if the poetry dependencies for all these scripts should be shared or not- fire away! I realize it makes the diff and PR size much scarier than it has to be. That is, I can make it one scripts poetry "project" and install all shared dependencies for all our dev-scripts instead.


api/Dockerfile
```diff
@@ -8,7 +8,8 @@ RUN pip install -r /api/requirements.txt
 COPY . /api
 WORKDIR /api
 
-RUN python -c "from sentence_transformers import SentenceTransformer; SentenceTransformer('all-mpnet-base-v2')"
+ENV TRANSFORMERS_CACHE="/api/big_models_data_cache"
+ENV TORCH_HOME="/api/big_models_data_cache"
```
ccjoel, line 12: These env vars are THE ONE IMPORTANT CHANGE on this diff, besides the `get_embedder` and `get_highlighter` changes below. Oh and the docker-compose volume shared setup.

You may ignore anything else under `api/scripts`, unless you're curious on the utility scripts we have to populate our database.

api/docker-compose.yaml
```diff
@@ -25,6 +25,7 @@ services:
         hard: -1
     volumes:
       - dojo_esdata:/usr/share/elasticsearch/data
+      - ./big_models_data_cache:/api/big_models_data_cache
```
ccjoel, line 28: Second most important change. We could decide to move to docker-compose-dev.yaml but I don't have enough information on how/if we use that yet.

api/src/documents.py
```diff
@@ -35,8 +35,8 @@
 from rq.exceptions import NoSuchJobError
 from rq import job
 
-from src.embedder_engine import embedder
-from src.semantic_highlighter import highlighter
+from src.embedder_engine import get_embedder
```
ccjoel, line 38: Other important changes! Retrieves singleton instance (or initializes it the first time).

api/docker-compose.yaml
ccjoel: @mattprintz I'm hoping you have both context on the docker-compose-dev file as well as opinion on if we should only use this caching mechanism on dev vs always (including prod). I'm of the opinion that we should do it both during local dev and in deployments, but could be persuaded otherwise.

mattprintz: docker-compose-dev isn't really used anywhere any longer. For "dev" purposes, the changes should be in the main `docker-compose.build-override.yaml` file.

The "local" docker-compose files (the ones in the sub directories) are the docker compose rules if we were going to be using docker compose to run just that service by itself, like if we had a multi-server deploy with api on one server and terminal on another, etc.

mattprintz: I think this change is fine for both prod and dev.

mattprintz (approved):

ccjoel:
Unresolved items on this PR:

1. What happens if multiple searches come in while the embedder is loading the first time? Does that result in multiple, simultaneous loading?

2. [...] look in to having the embedder hot load asyncronously after api startup. Seems like the would be the best of both worlds if get_embedder returned an async promise [...] might be an idea worth considering?

ccjoel:
Closing to split between scripts cleanup / and LLM-JIT loading.

PR #111, opened 2023-06-09
PCA Weights Alternatives from Research

ccjoel:
This refactors and splits the pca->weights causemos_processors job toadd 3 different pca-to-weight algorithms as well as add some tests to visualize what the various steps do.

# Done:
- Add endpoint/job context param to tweak v3's threshold variance explains by selected PCs so that it can be overwritten instead of hardcoded to 0.7  (70%)

## How I tested Manually
```
python tasks/causemos_processors.py
```

Which uses the original nd-gain.json from sample index model and logs several steps as well as the final output.
Basically a full integrated manual test.

## Running Automated Tests

```
python -m pytest -vvs tasks/causemos_processors_test.py
```

Which will perform various tests for intermediary steps, instead of the complete integrated work.


tasks/tasks/causemos_processors.py
```diff
@@ -140,35 +223,156 @@ def generate_index_model_weights(
...
+                weights_row = weight_matrix[pc_index]
+                weights_row[feature_index] = weights_row[feature_index] + relative_value
+
+    matrix_percents = list(map(calc_percents, weight_matrix))
+    matrix_averages = [(sum(v)/amount_relevant) for i, v in enumerate(zip(*matrix_percents))]
+
+    return matrix_averages
+
+
+def pca_to_weights_v3(pca_details: Tuple, target_pca_explained_sum=0.7):
```
ccjoel, line 283: @brandomr this is the algorithm you shared on slack and we went through recently. It is the one wired to the whole flow as of the current state of this PR.

I added the other v1, v2 version functions in order for us to test/toggle and compare as well.

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+    first_dataset_feature = first_dataset_stored["source_column"]
+    first_dataset_feature_name = first_dataset_stored["name_in_model"]
+
+    assert first_dataset_feature == "governance"
+    assert first_dataset_feature_name == "Governance indicator"
+
+@pytest.mark.integration
+@pytest.mark.side_effects
+def test_download_datasets():
+    out = download_datasets(iteration_func(index_model_object))
```
ccjoel: Actually don't run this tests unless you want to really download the datasets from dojo. I think I copied the files over since I already had them! Forgot about this. I may remove this "test" entirely before merging this work (testing that files download is odd!) but it helped me during development.

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+        [0.0011, 0.0401, 0.7065, 0.7065, 0.0],
+        [0.7139, 0.6997, 0.0193, 0.0193, 0.0],
+        [0.7002, 0.7133, 0.0208, 0.0208, 0.0],
+        [0.0, 0.0, 0.7071, 0.7071, 0.0],
+        [0.0, 0.0, 0.0, 0.0, 1.0]
+    ]
+
+
+@pytest.mark.integration
+def test_pca_to_weights_v3_70_control_data():
```
ccjoel, line 299: @brandomr this is me running the v3 pca-to-weight algorithm shared on slack. The pca details input is what the `perform_pca` would output for a dataset with 1000 rows and those 5 columns. This is what I call control data/set.

It has the threshold of 70% set for explained cumulative variance from selected PCs.

Check the output (assert of the rounded output) below

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+            [5.00405570e-01, 2.60246340e-01, 2.39348091e-01,
+             2.51310581e-32, 0.00000000e+00]
+        )
+    )
+
+    weights = pca_to_weights_v3(pca_details, 0.9)
+
+    rounded = list(map(lambda i: roundToFour(i), weights.tolist()))
+
+    assert rounded == [0.2434, 0.2565, 0.25, 0.25, 0.0]
```
ccjoel: @brandomr this is the updated algorithm with the control data, with cumulative pca explained variance set to 90%.

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+
+    assert sum(rounded) > 0.99
+
+
+@pytest.mark.integration
+def test_pca_to_weights_v3_70_explained_sum_real_index_datasets():
```
ccjoel, line 363: @brandomr this is the v3 pca-to-weight with 70% explained variance PCs, with the real index ndgain datasets.

Oh and I realize the output of these tests are formatted as actual percentages, but should be easy to compare to the other results (which were not multiplied by 100)

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+
+    weights = pca_to_weights_v3(pca_details, target_pca_explained_sum=0.9)
+
+    rounded_as_percent = list(map(lambda i: f"{roundToFour(i) * 100}%", weights.tolist()))
+
+    assert rounded_as_percent == ['4.73%', '5.7700000000000005%', '5.62%', '5.8999999999999995%', '4.07%', '5.66%', '0.0%', '4.65%', '5.47%', '5.140000000000001%', '5.0%', '3.53%', '4.4799999999999995%', '4.47%', '4.68%', '4.91%', '5.029999999999999%', '4.8500000000000005%', '4.9799999999999995%', '5.45%', '5.609999999999999%']
+
+
+@pytest.mark.integration
+def test_pca_to_weights_v2_control_data():
```
ccjoel: @brandomr this is the v2 "updated matrix" (hah) algorithm I had worked before, running against the control data (pca analysis output from the same control data as the other tests)

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+    weights = pca_to_weights_v2(pca_details)
+
+    rounded_as_percent = list(map(lambda i: f"{roundToFour(i) * 100}%", weights))
+
+    assert rounded_as_percent == ['32.45%', '33.32%', '17.11%', '17.11%', '0.0%']
+
+
+
+@pytest.mark.integration
+def test_pca_to_weights_v2_real_index_data():
```
ccjoel, line 417: @brandomr this is the v2 "updated matrix" (hah) algorithm I had worked before, running against the real ndgain dataset output after performing pca on it.

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+
+    weights = pca_to_weights_v2(pca_details)
+
+    rounded_as_percent = list(map(lambda i: f"{roundToFour(i) * 100}%", weights))
+
+    assert rounded_as_percent == ['2.76%', '3.88%', '3.65%', '5.82%', '5.53%', '5.8500000000000005%', '0.0%', '3.6799999999999997%', '6.7299999999999995%', '5.35%', '6.63%', '5.33%', '3.44%', '3.4000000000000004%', '6.18%', '6.09%', '6.99%', '6.13%', '5.55%', '3.2800000000000002%', '3.71%']
+
+
+@pytest.mark.integration
+def test_pca_to_weights_v1_control_data():
```
ccjoel, line 433: @brandomr: This is the original pca-to-weight v1 algorithm which is currently deployed (not updated matrix one nor the one you suggested in slack). It's running aginst pca output for the control dataset.

The other following test is the same v1 nbut running against the real nd gain.

tasks/tasks/causemos_processors_test.py
```diff
@@ -0,0 +1,415 @@
...
+    weights = pca_to_weights_v2(pca_details)
+
+    rounded_as_percent = list(map(lambda i: f"{roundToFour(i) * 100}%", weights))
+
+    assert rounded_as_percent == ['32.45%', '33.32%', '17.11%', '17.11%', '0.0%']
+
+
+
+@pytest.mark.integration
+def test_pca_to_weights_v2_control_data():
```
ccjoel, line 386: This is a duplicate test, re-pasted by mistake. Ignore, as I'll remove later.

brandomr:
Looks great! For simplicity and clarity I'd recommend removing the alternate PCA approaches; we can reimplement them off prior commits later if need be.

PR #101, opened 2023-06-05
Data Transformation: Geo Regrid Aggregation Function Selector

adamgilg:
This adds a select for the geo regrid aggregation function on the Data Transformations page. At the time of this PR, the backend side of things is not hooked up, and the link to `https://www.dojo-modeling.com/data-registration.html#adjust-geospatial-resolution` just takes you to the general data registration docs page. 

There's a PR for the docs update, and the backend work is in progress. 

When the backend work is done, the arguments sent to the API may need to change slightly - the API call is modeled after the temporal regridding call, which takes an array with a single aggregation function. We'll revisit this once the endpoint is done to test it. 

Screenshot 2023-06-05 at 1 24 13 PM

Screenshot 2023-06-05 at 1 24 07 PM

PR #116, opened 2023-06-23
Geo & Temporal Resolution Data Transformation Updates

adamgilg:
This PR is currently blocked awaiting the Elwood changes to enable Geo Resolution to work with non-uniform data. 

It also may be merged with @ccjoel 's work on the Data Transformation step. Further description to come.


ccjoel:
Quick note that the dataset cypress tests are failing. That's expected, as we're making the transformations step required and the tests don't account for that. I'll address it by adding mock job responses that indicate to the UI that no action is needed (whatever that means- it can be contradicting info (no lat/lon nor country annotations) as we only need that mock to allow the UI to proceed through the transform step all the way to preview->submit.

Not extremely sensitive nor exactly required to merge our work. I'll see what I can do.

ui/client/datasets/DataTransformation/AdjustGeoResolution.js
```diff
@@ -70,47 +100,93 @@ export default withStyles((theme) => ({
...
+      if (validNumber(selectedResolution)) {
+        return `${selectedResolution} km`;
+      }
+
+      return '';
+    }
+    return `${selectedResolution.toFixed(2)} km`;
+  };
+
+  const disableButton = () => {
```
ccjoel, line 187: I don't think I understand why this is a function and not a boolean. Would be helpful if you leave a note on the code, or maybe we can talk through it here.

eg what I have in my mind:

```
const areTransformButtonsDisabled = (nonUniform && !validNumber(selectedResolution)) || !selectedResolution;
```

ccjoel:
(thanks for sharing- I'll let you know when I go back to the gadm-resolver work and we can sync then)

adamgilg:
I manually brought in the changes from https://github.com/jataware/dojo/pull/101 and did some updating to make them work with all the new `non-uniform/event data` stuff. That PR is now closed.

brandomr:
blocked by https://github.com/jataware/elwood/pull/16

adamgilg:
> blocked by [jataware/elwood#16](https://github.com/jataware/elwood/pull/16)

No longer blocked.

tasks/tasks/resolution_processors.py
```diff
@@ -59,19 +59,25 @@ def calculate_geographical_resolution(context, filename=None, **kwargs):
 
     resolution_res = detect_latlon_resolution(lat, lon)
 
-    if resolution_res is None or resolution_res.square is None:
+    if (
+        resolution_res is None
+        or isinstance(resolution_res, dict)
+        or resolution_res.square is None
+    ):
+        print(f"RESOLUTION RESULT: {resolution_res}")
```
ccjoel, line 67: Quick question- do you get this printed as output?
I've seen it print in some systems or circumstances, but not on others.
What I ended up landing on to get consistent prints is to import and using the logging (example setup up: https://github.com/jataware/dojo/blob/main/tasks/tasks/causemos_processors.py#L27)

Just in case any of this is helpful!

ui/client/datasets/DataTransformation/DataTransformation.js
```diff
@@ -181,6 +189,60 @@ const DataTransformation = withStyles(() => ({
   // eslint-disable-next-line no-unused-vars
   const [failedJobs, setFailedJobs] = useState([]);
 
+  // toggles when the RunTransformations overlay closes
+  const [allTransformationsComplete, setAllTransformationsComplete] = useState(false);
+  // toggles when the RunTransformation component unmounts due to errors
+  const [transformationProcessingError, setTransformationProcessingError] = useState(false);
+
+  // Handles the next step process after the transformations are complete in <RunTransformation>
+  useEffect(() => {
```
ccjoel, line 198: I don't follow this useEffect section, as there is a different `handleNextStep` that we need to call. Not sure what `handleNext` is in this context. 😕

PR #117, opened 2023-06-23
Added new hybrid search implementation

ccjoel:
New implementation does not use fuzzy matching on the keyword match side, as well as consolidates results of both keyword and semantic on the python server, by alternating results 3 at a time.

How it looks integrated with the UI:
Screenshot from 2023-06-23 16-18-14

Keyword search usually has higher match % than semantic. Does the back and forth between colors and match% bar sizes bother anyone else? See screenshot above. Got feedback on this and punted for later, as this dojo interface is not targeted for external users, but for debugging/tweaking search purposes.

### DIFF
--- a/api/scripts/hybrid_search.py
+++ b/api/scripts/hybrid_search.py
@@ -78,12 +78,10 @@ QUERIES = [
     hybrid_query_v1,
     keyword_query_v1,
     hybrid_query_v0,
-    # hybrid_query_v2,
+    hybrid_query_v2,
     keyword_query_v3
 ]
 
-# TODO expand so that query is queries (list) and we call all of them if needed
-#      then also receive an additional post_op function that applies some operation to the results of all queries
 def search(query):
 
     # NOTE this is all hacked up but Ok for now; args.size would fail
--- a/api/scripts/hybrid_search.py
+++ b/api/scripts/hybrid_search.py
@@ -42,7 +42,9 @@ from src.feature_queries import (
     semantic_search_query,
     hybrid_query_v1,
     keyword_query_v1,
-    hybrid_query_v0
+    hybrid_query_v0,
+    hybrid_query_v2,
+    keyword_query_v3,
 )
 
 import collections
@@ -61,7 +63,7 @@ parser.add_argument("-q", "--query",
 
 parser.add_argument("-s", "--size",
                     help="es scroll page result size",
-                    type=int, default=10)
+                    type=int, default=100)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


api/scripts/hybrid_search.py
```diff
@@ -42,7 +42,9 @@
     semantic_search_query,
     hybrid_query_v1,
     keyword_query_v1,
-    hybrid_query_v0
```
ccjoel, line 45: This is the query-comparison-report helper script. Feel free to ignore.

api/src/feature_queries.py
```diff
@@ -1,5 +1,36 @@
...
+def keyword_query_v3(phrase):
+    """
+    """
+    q = {
+        "query": {
+            "bool": {
+                "should": [
+                    {
+                        "match_phrase": {
+                            "name": phrase
```
ccjoel: New keyword query shared on slack. Without fuzzy matching. Used together with our semantic search implementation for the latest hybrid search.

api/src/indicators.py
```diff
@@ -137,38 +137,45 @@ def patch_indicator(
...
+    hit["matched_queries"] = ["semantic_search"]
+    return hit
+
+
 @router.get(
     "/features/search", response_model=IndicatorSchema.FeaturesSemanticSearchSchema
 )
 def semantic_search_features(
-    query: Optional[str], size=10, scroll_id: Optional[str] = None
+    query: str, size=10 # , scroll_id: Optional[str] = None
```
ccjoel, line 153: Can't receive or return a scroll id- which scroll_id should we use if we are merging the results of two queries manually? I don't think anyone was scrolling with the semantic search endpoint (dojo wasn't), but we can confirm with Uncharted.

api/src/utils.py
```diff
@@ -301,3 +301,42 @@ def add_date_to_dataset(path):
...
+    stepping by 3. Example output:
+
+    ["k1", "k2", "k3", "s1", "s2", "s3", "k4", "k5", "k6", "s4"]
+
+    If id_property is provided, it is presumed the lists contain dictionaries,
+    and the duplicates are removed by checking the id_property in the dictionary
+    """
+
+    if id_property:
+        list2 = get_unique_items(list1, list2)
```
ccjoel, line 328: TODO add id_property here, since this is working only because get_unique_items is defaulting to the same attribute (see its fn signature).

ccjoel: Second optional TODO might be to also allow to remove duplicates when not a list of dicts. Not part of actual usage, but would help get_unique_items be more generic and align better with the tests I added.

ui/client/components/ViewFeatures.js
```diff
@@ -364,7 +364,6 @@ const ViewFeatures = withStyles(() => ({
           LoadingOverlay: CustomLoadingOverlay
         }}
         loading={featuresLoading}
-        getRowId={(row) => `${row.owner_dataset.id}-${row.name}`}
```
ccjoel, line 367: This was an old addition, not necessary anymore as the features include an id, which DataGrid uses by default. The existing feature IDs are exactly this (a concatenation of their owner dataset and their name).

PR #118, opened 2023-06-29
Gadm resolver all

ccjoel:
Draft of  changes for gadm resolver, compared to move-temporal-resolution (which is merged into this feature branch).

Pending:
- How/when to merge all transform changes with this, and merge order.


ui/client/datasets/DataTransformation/GadmResolver.js
```diff
@@ -0,0 +1,275 @@
...
+      return acc;
+    }, {});
+
+    onSave({ [primaryField]: formatted });
+  };
+
+  return (
+    <div
+      className={classes.root}
+      style={{ minHeight: 500 }}
```
adamgilg, line 213: This minHeight could be moved into the `root` class

ui/client/datasets/DataTransformation/DataTransformation.js
```diff
@@ -257,6 +311,28 @@ const DataTransformation = withStyles(() => ({
...
 
+  const onGadmResSuccess = useCallback((resp, setData, setDataError, setDataLoading) => {
+    if (resp) {
+      if (resp.fuzzy_match && resp.fuzzy_match.length) {
+        setData(resp);
+      } else {
+        setDataError('Nothing to review.');
+      }
+    } else {
+      setDataError('Something went wrong. Please contact Jataware for assitance.');
```
adamgilg, line 323: Typo: assitance -> assistance (both here and on line 331)

ui/client/datasets/DataTransformation/useElwoodData.js
```diff
@@ -64,7 +65,27 @@ const useElwoodData = ({
       };
 
       if (datasetId) {
-        startElwoodJob({ requestArgs }).then((jobId) => repeatFetch(jobId));
+        startElwoodJob({ requestArgs })
```
adamgilg, line 68: I'm going to pull this useElwoodData change into a new branch to start the work fixing the repeat call flow we talked about last week.

adamgilg:
This looks good! I haven't pulled it and tried it out locally yet but I'll give that a shot when it's ready for proper review.

ui/client/datasets/DataTransformation/useElwoodData.js
adamgilg: Actually, I'm just going to bring in a skeleton version of this to avoid merge conflicts.

ui/client/datasets/DataTransformation/useElwoodData.js
```diff
@@ -9,6 +9,7 @@ const useElwoodData = ({
   generateArgs,
   jobString,
   cleanupRef,
+  onBackendFailure
```
adamgilg, line 12: `onBackendFailure` needs to be added as a dependency to the useEffect. I disabled the linter for the dependency array here because I didn't think I should be adding the ref to it, but on further exploration I believe it should be totally safe to add the ref without triggering any extra useEffect updates. So we can remove  `eslint-disable-next-line react-hooks/exhaustive-deps` from line 116 and add `cleanupRef` (and `onBackendFailure`).

ccjoel:
Closed, as created a new PR based against main.

PR #122, opened 2023-07-07
Update to Transformation Page to account for already running/finished jobs

adamgilg:
This PR refactors the useElwoodData hook to handle Elwood transformation jobs that are already running or have already finished.

### DIFF
--- a/ui/client/datasets/DataTransformation/useElwoodData.js
+++ b/ui/client/datasets/DataTransformation/useElwoodData.js
@@ -16,61 +16,110 @@ const useElwoodData = ({
   const [dataError, setDataError] = useState(false);
 
   useEffect(() => {
+    // Kicks off the elwood job
     const startElwoodJob = async ({ requestArgs }) => {
-      console.log('these are the requestArgs when starting ', jobString, requestArgs);
       const jobQueueResp = await axios.post(
         `/api/dojo/job/${datasetId}/${jobString}`, requestArgs
       );
 
       if (jobQueueResp.status === 200) {
-        return jobQueueResp.data?.id;
+        return jobQueueResp.data;
       }
     };
 
-    const runElwoodJob = ({ requestArgs, onFailure }) => {
-      let count = 0;
-
-      const repeatFetch = (jobId) => {
-        setTimeout(() => {
-          axios.post(`/api/dojo/job/fetch/${jobId}`).then((response) => {
-            if (response.status === 200) {
-              // keep track of how long it takes (for dev purposes)
-              count += 1;
-              // console.log(`${count}: response from job string: ${jobString}:`, response);
-              if (response.data) {
-                console.log(`success! ${jobString} took ${count * 500}ms`, response.data);
-                // setOptions currently just used for onGeoResSuccess
-                onSuccess(response.data, setData, setDataError, setDataLoading, setOptions);
-                return;
-              }
-              if (cleanupRef.current) {
-                // if no data and the component is mounted, try the fetch again
-                repeatFetch(jobId);
-              }
+    // repeatedly calls /job/fetch/jobId until it gets the job results
+    const repeatFetch = (jobId, requestArgs, onFailure) => {
+      setTimeout(() => {
+        axios.post(`/api/dojo/job/fetch/${jobId}`).then((response) => {
+          if (response.status === 200) {
+            if (response.data) {
+              // setOptions currently just used for onGeoResSuccess
+              onSuccess(response.data, setData, setDataError, setDataLoading, setOptions);
+              return;
             }
-          }).catch((err) => {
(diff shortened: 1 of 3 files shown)


ui/client/datasets/DataTransformation/useElwoodData.js
```diff
@@ -16,61 +16,110 @@ const useElwoodData = ({
...
+        axios.post(`/api/dojo/job/fetch/${jobId}`).then((response) => {
+          if (response.status === 200) {
+            if (response.data) {
+              // setOptions currently just used for onGeoResSuccess
+              onSuccess(response.data, setData, setDataError, setDataLoading, setOptions);
+              return;
             }
-          }).catch((err) => {
-            // we get a 404 immediately - there is some sort of bug, this accounts for that
-            if (count < 2 && cleanupRef.current) {
```
adamgilg, line 54: I believe the bug this comment references was due to my very slow system/docker setup on my old computer. I would get a 404 back on the first `/dojo/job/fetch/${jobId}` call every time. This isn't happening anymore, so the `count` check doesn't seem necessary.

ui/client/datasets/DataTransformation/useElwoodData.js
```diff
@@ -16,61 +16,110 @@ const useElwoodData = ({
...
+        });
+      }, 500);
+    };
+
+    // the jobId is always this format so we can create it here
+    const jobId = `${datasetId}_${jobString}`;
+    // check if the job is already running or finished when we first hit the page
+    const checkExistingJob = async () => {
+      try {
+        const response = await axios.post(`/api/dojo/job/fetch/${jobId}`);
```
adamgilg, line 74: We now will see an initial 404 the first time we hit the transformation page, however. That's because we need to check to see if the job is already running, and the very first time we get there it won't be running, so we'll get a 404 back, which tells us to kick it off. This is expected.

ccjoel:
Hm. There's an edge case to the reload situation (not related to the solution here), that I hadn't realized until now!

I'll outline it here so that we can think about it:

If a user annotates a dataset, proceeds and starts the transformation-fetch jobs, then pressed back to the annotations page and changes the annotations quickly, navigating forward again to the transformations page will use the cached response with the wrong, previous, annotations.

I don't think know if this would be a common scenario. The rqworker jobs results are stored for 500 seconds (8 minutes). If the user annotates, proceeds to transform, finishes the flow and goes back to modify the annotation after registering the dataset, 8 minutes will likely have gone by and the jobs will run again upon changing the annotations.

If we wanted to be completely sure that a job should not be re-run, we would likely need to involve a datetime or hash on the annotation data to help the UI or backend decide if the previous job results, if available, should be used.

I'll think about this for a little bit more. Besides that, the changes here work as expected 🤖

ccjoel:
We discussed my last comment/concern and agreed that the Dojo UI should issue a "clear transformations fetch job results" anytime it updates annotations from the annotate step/page. This is pending for this work to be ready.

adamgilg:
> We discussed my last comment/concern and agreed that the Dojo UI should issue a "clear transformations fetch job results" anytime it updates annotations from the annotate step/page. This is pending for this work to be ready.

@ccjoel https://github.com/jataware/dojo/pull/122/commits/1c4af1eb30d0612e39050f3069ba043b905f1db5 should resolve this issue. I added a `/job/clear/{uuid}` endpoint that clears rqworker jobs for a uuid and set it up to be called when you hit `Next` on the Annotations page.

ccjoel:
I just realized this work is ready for review. I thought it was waiting on the other blocked PR. I'll confirm with Adam and review ASAP.

adamgilg:
I pulled in the latest changes and modified useElwoodData to work with the onBackendFailure PromptDialog, now accepting multiple failures. You can see a screenshot below. 

Screenshot 2023-07-25 at 1 55 08 PM

ccjoel (approved):
I tried them out. Looks good to me!

ccjoel:
There were some test failures, maybe regarding mocking/stubbing the clear endpoint or something else? Test shouldn't fail if we just call the clear job endpoint and that fails, so I went ahead and merged. I think it is fine to address once we're done upgrading the cypress version on the other PR.

ccjoel:
> Test shouldn't fail if we just call the clear job endpoint and that fails, so I went ahead and merged

Lol... as I discovered on the cypress upgrade work, tests WERE FAILING because the call to job/clear failing. I changed the cypress setting to not do that anymore. See:

https://github.com/jataware/dojo/blob/main/ui/cypress/support/e2e.js#L19

(
```
Cypress.on('uncaught:exception', (err, runnable) => {
  // If an inconsequential http request is done and not mocked
  // ignore and continue test without failing
  return false;
});
```
)

We can revisit if this becomes a problem if we author new end-to-end tests with cypress.

PR #134, opened 2023-07-24
Bugfix/elwood processors filetype fix

Sorrento110:
Change the filetype passed to elwood from being hardcoded to being dynamically set from the context metadata.

### DIFF
--- a/tasks/tasks/elwood_processors.py
+++ b/tasks/tasks/elwood_processors.py
@@ -25,7 +25,6 @@ from settings import settings
 logging.basicConfig()
 logging.getLogger().setLevel(logging.DEBUG)
 
-
 def build_elwood_meta_from_context(context, filename=None):
     metadata = context["annotations"]["metadata"]
     if "files" in metadata:
@@ -78,7 +77,7 @@ def dict_get(nested, path, fallback=None):
     if not type(nested) == dict:
         return fallback
 
-    keys = path.split(".")
+    keys = path.split('.')
     value = nested
     try:
         for key in keys:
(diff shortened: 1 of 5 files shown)


brandomr:
Pending my testing this with a real (date included) netcdf file produced by a model and then testing execution in airflow.

tasks/tasks/elwood_processors.py
```diff
@@ -441,7 +446,34 @@ def run_model_elwood(context, *args, **kwargs):
...
+        metadata = metadata["files"][filename]
+    
+    if 'filetype' in metadata:
+        ftype = metadata['filetype']        
+    elif 'filename' in metadata:
+        ftype = metadata['filename'].split('.')[-1]
+    else:
+        ftype = 'csv'
+    
+    if ftype == 'nc':
```
ccjoel, line 466: This will also fail for `nc4` file extensions, per our discussion yesterday (I think).

I would recommend creating a util function that properly does this logic, grouping multiple file extension variants per ftype (I think excel might have a different one for old files, as an example), so that we don't need to repeat this logic inline on rqworker/elwood independently.

tasks/tasks/elwood_processors.py
```diff
@@ -441,7 +446,34 @@ def run_model_elwood(context, *args, **kwargs):
...
+    elif ftype == 'tiff':
+        ftype = 'geotiff'
+    elif ftype == 'tif':
+        ftype = 'geotiff'
+    elif ftype == 'xlsx':
+        ftype = 'excel'
+
+    # TODO: this is a big hack to allow the preview to continue
+    # in model output annotation to use the raw csv
     mm_ready_annotations["meta"] = {"ftype": "csv"}
```
ccjoel, line 477: Actually this code looks outdated- should we close this PR?

ccjoel:
Closing as stale (+ conflicts). We'll revisit the  relevant portion if we need it later.

PR #152, opened 2023-08-11
Cypress: End-to-end and Integration tests, framework

ccjoel:
Needs some cleanup, but can get help on this after merging.

Follow-up PRs will come after, hopefully with help from others.
TODO: modify some hardcoded values for seeds from using localhost:8080 to using `Cypress.config('baseUrl')`, add ability to login with our stage auth. This is in order to run this tests against a deployed staged env instead of running E2E tests against local only.


ccjoel:
The exludeSpecPattern didn't kick in. Maybe because this branch doesn't have latest main merged in. Will retry soon.

.github/workflows/int_tests_cypress.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         uses: cypress-io/github-action@v5
         with:
           working-directory: ./ui
-          config: video=false,videoUploadOnPasses=false,keystrokeDelay=1,retries=2
+          config: '{"e2e": { "excludeSpecPattern": ["*.hot-update.js", "**/*.e2e.cy.js"] }, "video": false, "keystrokeDelay": 1, "retries": 2,  "component": {"excludeSpecPattern": ["*.hot-update.js", "**/*.e2e.cy.js"]}}'
```
ccjoel, line 39: Couldn't get it to work as before (inline config).. had to repalce with JSON. We'll clean up if we find an alternative, but this was blocking this PR.

ui/.eslintrc.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   "parser": "@babel/eslint-parser",
-  "extends": ["airbnb", "plugin:react-hooks/recommended", "plugin:storybook/recommended"],
+  "extends": ["airbnb", "plugin:react-hooks/recommended",
+              "plugin:storybook/recommended", "plugin:cypress/recommended"],
```
ccjoel, line 4: We still don't have a cypress lint task. Follow-up work to add it and address problems!

ui/client/components/templater/TextAnnotater.js
```diff
@@ -87,6 +87,7 @@ const Mark = ({
   const displayInnerContent = () => (
     <span
       className={savedHover ? classes.darkerMark : classes.mark}
+      data-testid="annotation-inner-contents"
```
ccjoel, line 90: This is testid and not test, as it's not a custom `cy.get(''[data-test=...)` and instead it is a built-in plugin: `cy.findByTestId(...)`. We'll consolidate in follow-up work as needed.

ui/client/datasets/RunJobs.js
```diff
@@ -122,7 +122,7 @@ const RunJobs = withStyles(({ spacing }) => ({
       console.log('failed');
     } else {
       // No result, wait for an update
-      timeoutHandle = setTimeout(updateJobData, 500);
+      timeoutHandle = setTimeout(updateJobData, 1500);
```
ccjoel, line 125: This makes the user wait an additional second once a job has finished, but helps both the tests and the deployment instance not go crazy memory-wise.

ui/cypress/e2e/README.md
```diff
@@ -0,0 +1,31 @@
+
+Depends on host:
+cy.onlyOn('localhost')
+
+onlyOn('mac', () => {
```
ccjoel, line 5: Argh, was looking for this earlier in the docs, couldn't find it. I'll add a proper helper README as follow-up work. Help needed.

ui/cypress/e2e/test_model.e2e.cy.js
```diff
@@ -0,0 +1,243 @@
...
+        cy.reload();
+
+        cy.findByRole('textbox', {name: /terminal input/i}).as('TerminalInput');
+
+        cy.wait(500);
+
+        cy.get('@TerminalInput')
+          .type(`sudo apt update && sudo apt install -y python3-pip && git clone https://github.com/jataware/test-model.git && cd test-model{enter}`, {force: true});
+
+        cy.wait(25000); // 60000 for slower computers
```
ccjoel, line 88: This has a bunch of overlap with `model.e2e.cy.js` (not missing `test_` prefix on file). This particular file tests registering our jataware/test-model, which means we have to type on the virtual xterm terminal and wait for long commands to finish (sudo apt update; sudo apt install ...; pip install ...). I don't know of a proper way to verify once a command in this ssh-virtual terminal is completed, so we have arbitrary waits here. You'll have to change it in order to work on your system. This test is a bit overkill, but really auto-tests registering the test model. I still recommend running all the model.e2e.cy.js tests (and component tests!) as it really tests all aspects of model registration piecemeal.

ccjoel: Help needed in order to check when the last running command has completed, without waits?

Maybe:

```
cy.get('@TerminalInput')
// select ^ this element, but with a not:disabled check, and add a very long timeout for it? Not great, but maybe better.
```

ui/cypress/plugins/index.js
```diff
@@ -27,5 +150,78 @@ module.exports = (on, config) => {
     );
   }
 
+  on('task', {
+
+    'debug': (...args) => {debug(...args); return true;},
```
ccjoel, line 155: This is to get access to this really nice node-console debug on tests and on when running on github!

Use like:

```
cy.task('debug', myVar);
```

Where `myVar` can be any param allowed to `console.log`.

Locally, this will be printed on the terminal log for cypress, like:

Screenshot from 2023-08-18 11-48-04

and on github action jobs:

Screenshot from 2023-08-18 11-48-52

ui/cypress/support/commands.js
```diff
@@ -80,3 +83,35 @@ Cypress.Commands.add('setSelection', { prevSubject: true }, (subject, query, end
     }
   })
 ));
+
+
+// let sharedStatus = 0;
+
+// Cypress.Commands.add('spyPoll', {}, (method, url, property, expectedVal, nth=0) => {
```
ccjoel, line 90: Previous attemp to loop and wait until, for example, an elwood job completed.

Freakin nightmare. I opted to use a separate helpers.js function with async/await and do `cy.request()`s, then use `cy.wrap()` to wait for this promise in tests.

Help needed to consider replacing the `cy.request()` to `axios` and internal helpers.js `cy.wait()` to a `sleep()` function that uses promises, maybe.

ui/cypress/plugins/index.js
```diff
@@ -27,5 +150,78 @@ module.exports = (on, config) => {
...
+
+    // TODO this is unused for now as we can use API
+    // for simple cases, but for complex scenarios it will be useful
+    'seed': ({type, id, name}) => {
```
ccjoel, line 159: My vision was to use `cy.task('seed'...)` everywhere on E2E tests so as to abstract away this chore, but didn't work as I wanted and is kinda a nice-to-have. Help needed. That is, change the test calls to seed from `cy.request(datasetUrl, data)... manual seed` to using `cy.task('seed', {type: 'dataset', ...)`

ui/cypress/seeds/model_api_data.js
```diff
@@ -0,0 +1,250 @@
...
+      // Municipality/Town=
+      "admin3": faker.helpers.multiple(faker.location.street)
+    },
+    "period": {
+      "gte": 0,
+      "lte": 10000000000
+    },
+    // "next_version": "string", // would mean the current version is deprecated in favor of next version
+    // "prev_version": "string",
+    "is_published": true,
```
ccjoel, line 176: Mocking this as published for the model list page, but then caught me by surprise (badly!) when reusing for the actual register tests (as the model was already "published" with this flag.).

TODO: create a second model seed mock fn based on this, maybe called `genUnregisteredModel`, and use that in the appropriate tests. Help needed.

PR #166, opened 2023-09-07
Upgrade React to 18.2.0

adamgilg:
This upgrades React and ReactDOM to 18.2.0, the latest release (from summer 2022). 
A few libraries we used required upgrades as well, as they didn't support React 18:

- react-leaflet: v3 -> v4, which required some small code changes in the `ClipMap` component due to removal of a prop we were using
- cypress/react: there was a hidden cypress/react18 library that isn't documented anywhere, just mentioned in a closed issue. This appears to be the way they're supporting React 18.
- @mui/styles: removed. This is just included in mui v5 to support the legacy makeStyles/withStyles API. Due to changes in React 18, [this won't work anymore](https://github.com/mui/material-ui/issues/32142#issuecomment-1289529626), which is why we're now using `tss-react`.
- react-hot-loader: removed. React Hot Loader is not supported in React 18, and while Fast Refresh (the React Native solution) seems like it may work with some configuring, our hot loading seems to still work with this removed. Are we already using something else built into webpack? Not 100% sure, but removing this doesn't seem to cause any issues with development. 
- overrides: I vaguely remember adding this react-leaflet/ react-leaflet/core override, but I have no idea why. Removing it doesn't seem to cause any issues. 

Also: after upgrading to React 18, RunJobs.js was throwing errors that were preventing the dataset analysis and processing steps from running. I spent a bit of time refactoring it and came up with a more resilient (and React-happy) solution that still works to navigate back (which was the problem with just adding all the dependencies to the deps array of the useEffect in RunJobs). 

There are a number of other libraries that are currently throwing warnings about not supporting React 18, but the ones above were the only ones that were actually preventing the app from running. We can upgrade the rest as we see fit.


ccjoel:
Unfortunately upgrading storybook to work with React18 also included some nice demos that I liked and left in, which makes the diff a bit longer than before. Anything storybook related can be ignored and shouldn't really block merging of this PR, so I wouldn't worry much about it. Storybook is working on my system now.

ccjoel:
The Data Transform processing (loading) overlay lagged for me on Ubuntu. I couldn't take a screenshot because it finished quickly, but it looked like it was missing styles.

ccjoel (approved):

PR #169, opened 2023-10-02
Disables GADM resolution if no primary country is annotated

adamgilg:


ccjoel (changes requested):
Quickly reviewing my understanding:
I believe GADM resolution problem is related to the first one in transform step:

Screenshot from 2023-10-04 15-49-15

It only properly works when a `country` column is annotated as `primary`.
We likely need to change the logic on `generateFetchGadmArgs` to check if the primary column is annotated as `country`.

Let's sync, since I'm surprised that gadm resolution only works for you on lat/lon annotations and not on countries. There may be an edge case or bug involved. Easier to get on a call and figure it out. Thanks!

ccjoel:
Created #171 to handle this in the backend (rqworker) as well. That is, if the gadm resolution alternatives doesnt find a primary geo annotated, it returns `[]` for the fuzzy match result.

ccjoel (approved):
+1

PR #171, opened 2023-10-04
Handled case where no primary geo is set while running rqworker- gadm…

ccjoel:
… resolution_alternatives job.

This complements a UI fix where the job won't even get started in the first place, but should help make the system more stable.

### DIFF
--- a/tasks/tasks/gadm_processors.py
+++ b/tasks/tasks/gadm_processors.py
@@ -17,82 +17,26 @@ def resolution_alternatives(context, filename=None, **kwargs):
 
     admin_level = kwargs.get("admin_level", "country")
 
+    has_primary_geo = False
+
     geo_column = ""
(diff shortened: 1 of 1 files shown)


tasks/tasks/gadm_processors.py
```diff
@@ -17,83 +17,27 @@ def resolution_alternatives(context, filename=None, **kwargs):
...
+            "field": None,
+            "fuzzy_match": []
+        }
+
     # Run job
     gadm_response = elwood.get_gadm_matches(original_dataframe, geo_column, admin_level)
 
-    # Return hardcoded response in job
-    # provide gadm alternatives
-    mock_gadm_alternatives = {
```
ccjoel, line 34: Removed this old mock "sample" data (was unused), which was useful to illustrate the response body, but will easily get stale.

adamgilg (approved):

Issue #172, opened 2023-10-05
Fix dataset annotation multipart geo form 


adamgilg:
After the MUI upgrade, the multipart geo form on the dataset annotation page is slightly broken. It still works, but the label is overlaid on the input text for the main selected item (if started via `country`, country is broken, for example). All the other inputs work fine. There are also a number of MUI warnings that pop up in the console, these should be looked into. 

We're also making all of the multipart pieces required, but doing nothing to actually require them before saving the panel. We should make this actually required if we want to ensure that the user submits all four parts. 

![Screenshot 2023-10-05 at 11 10 16 AM](https://github.com/jataware/dojo/assets/2448578/f6a5d047-e3ea-4283-b92d-77eb676d5f77)

adamgilg:
This also applies to the text on the multipart date form, though the required validations work (better) on that one. 

I believe I know what the problem is with the visual issue. It shouldn't be a big fix.

PR #174, opened 2023-10-06
Dataset geo/date multipart annotation form fix

adamgilg:
https://github.com/jataware/dojo/issues/172

Both the geo & date multipart forms were looking bad (as was the model templater) because the labels weren't shrinking (becoming small at the top of the input). The label ended up overlaid on top of the initially selected input content, making it difficult to read. 

All the uses of FormikSelect - the Templater and both multipart forms - were not showing any `required` errors. This PR also fixes that. 

Screenshot 2023-10-06 at 11 55 21 AM

### DIFF
--- a/ui/client/components/formikComponents/FormikSelect.js
+++ b/ui/client/components/formikComponents/FormikSelect.js
@@ -3,8 +3,20 @@ import { useField, useFormikContext } from 'formik';
 
 import InputLabel from '@mui/material/InputLabel';
 import FormControl from '@mui/material/FormControl';
+import FormHelperText from '@mui/material/FormHelperText';
 import MenuItem from '@mui/material/MenuItem';
 import Select from '@mui/material/Select';
+
+// removes the [' '] from dataset multicolumn date & geo, which are there
+// because of the formik getIn selector & its difficulties with .nested names
+// but we need to match the name with the errors object name, which doesn't have the ['']
+function stripBrackets(str) {
+  if (str.startsWith("['") && str.endsWith("']")) {
+    return str.substring(2, str.length - 2);
+  }
+  // if it doesn't have the brackets, return the original string
+  return str;
+}
 
 /**
  * Wraps MUI Select for easier use with Formik. Needs to be within a Formik form.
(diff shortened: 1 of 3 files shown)


ccjoel:
Found a  very edgey and interesting edge case:

[Screencast from 2023-10-09 11-10-32.webm](https://github.com/jataware/dojo/assets/1967061/c1a8a896-4fc3-4022-b3b6-41fa455d62da)

Users may break the app if they explore too much. I've never seen a recursive/loop on this page/form/formik/mui like this. I wonder if the maximum callstack error was caused by the recent changes + changes on this PR.

I was able to reproduce consistently following exactly that order of operations from the video.

@adamgilg I'll let it up to you on if you think this is worth pursuing.

ccjoel:
I forgot to mention- this looks great btw. All of it went back to how it looked before the mui upgrade: the alignments of the fields, the alignment of the qualifies, the label shrink property 💯 !

ui/client/components/formikComponents/FormikSelect.js
```diff
@@ -3,9 +3,21 @@ import { useField, useFormikContext } from 'formik';
 
 import InputLabel from '@mui/material/InputLabel';
 import FormControl from '@mui/material/FormControl';
+import FormHelperText from '@mui/material/FormHelperText';
 import MenuItem from '@mui/material/MenuItem';
 import Select from '@mui/material/Select';
 
+// removes the [' '] from dataset multicolumn date & geo, which are there
```
ccjoel, line 10: Great comment. I was trying to find evidence of this workaround online, and I did find an old github issue that illustrates this. Will leave here in the PR for reference:

https://github.com/jaredpalmer/formik/issues/2278

ccjoel:
Tested it manually to ensure:
- Unit tests pass (build-a-get data is transform as it should). All UI jest unit tests passed.
- Checked that the input of the processing fn still works, and that all 3 portions of the build-a-date are sent to backend on next/save to Transformation page (overlaps somewhat with the unit test validity).

If we feel we can let the edge case I found for later, we can file an issue for it. Would be nice to address it- maybe timebox it? Let me know if you need any help with that. Thanks!

adamgilg:
> Found a very edgey and interesting edge case:
> Screencast.from.2023-10-09.11-10-32.webm
> 
> Users may break the app if they explore too much. I've never seen a recursive/loop on this page/form/formik/mui like this. I wonder if the maximum callstack error was caused by the recent changes + changes on this PR.
> 
> I was able to reproduce consistently following exactly that order of operations from the video.
> 
> @adamgilg I'll let it up to you on if you think this is worth pursuing.

I just checked, this bug also exists on `main` (that is, it isn't related to this fix). The issue, as I see it, is twofold: 

1. We're allowing users to select columns that have already been annotated in the various dropdowns for multicolumn annotations. Presumably something can't be annotated twice, and if it can, we'll need to ensure that this actually works. 
2. We need to make the date `FormatValidationInput` check run again when the select changes. That you're able to change the value and it doesn't display an error also contributes to this (and also exists on `main`). 

If we fixed both of these things, this wouldn't be possible. It is quite an edge case, however, so I don't really see it being worth the effort right now. But definitely something to file an issue for.

adamgilg:
> I forgot to mention- this looks great btw. All of it went back to how it looked before the mui upgrade: the alignments of the fields, the alignment of the qualifies, the label shrink property 💯 !

And the fields show errors now, which they didn't before!

adamgilg:
@ccjoel I opened https://github.com/jataware/dojo/issues/176 & https://github.com/jataware/dojo/issues/177 about the above.

ccjoel (approved):
🚢 ➡️ 🌎 ❗

PR #184, opened 2023-10-26
Data Modeling

adamgilg:
This PR adds the Data Modeling UI and backend + flowcast task.


docker-compose-images.yaml
```diff
@@ -0,0 +1,1159 @@
+name: tasks
```
ccjoel, line 1: Hm- Why was this file added here?

api/src/indicators.py
```diff
@@ -287,6 +287,46 @@ def get_latest_indicators(size=10000):
...
+            "deprecated",
+            "maintainer.name",
+            "maintainer.email",
+            "fileData.raw.url",
+            "fileData.raw.rawFileName",
+        ],
+        "query": {
+            "bool": {
+                "must": [
+                    {"match_all": {}},
```
ccjoel: Not that important and we already talked about it, but this query would work as well:
```
{
          "match": {
             "fileData.raw.url": ".nc"
          }
        }
```

(possibly same for rawFileName). Noting here that `must` makes it so that a result must match both *.raw.url and *raw.rawFileName (like an `and`, not `or`). Not sure if there were scenarios where some files had one or the other but not both, but mentioning it just in case.

docker-compose-images.yaml
ccjoel: (by mistake or is it necessary?)

ui/client/dagpipes/overview.css
```diff
@@ -0,0 +1,30 @@
...
+  border-radius: 2px;
+  border: 1px solid #f5f5f6;
+  &.selected {
+    border-color: #BBB;
+  }
+}
+
+.react-flow__node-sum,
+.react-flow__node-load,
+.react-flow__node-save, {
```
ccjoel, line 24: Doing a mental note on these, since the rest of styles are using makeStyles

tasks/tasks/data_modeling.py
```diff
@@ -0,0 +1,398 @@
...
+            "email": "todo@jataware2.com",
+            "organization": "Jataware2",
+            "website": "jataware.com"
+        },
+        "file_metadata": {
+            "filetype": "nc"
+        },
+        #TODO: this probably isn't the right way to set the filedata stuff 
+        # it looks like the endpoint is supposed to be setting it, but that wasn't working, while hardcoding it here did
+        "fileData": {
```
ccjoel, line 226: Yeah- I think this object and properties come from a misunderstanding I had when working on my first project here- porting spaceTag to Dojo. You'll notice fileData, rawFileName use camelcase, since the UI js code is setting that on its "internal" state, which is added to a dataset object, and then sent to the server to be merged (patched) onto the indicator object onto DB (es). I'm not sure if we had all the necessary info before fileData was added. Maybe we did, or maybe it needed to be set by the API when creating indicators.

For now it is out of scope to address, and Ok to do this manually here to create consistency between indicators created using the UI flow and this task.

tasks/tasks/data_modeling.py
```diff
@@ -0,0 +1,398 @@
...
+    #get datatype of data
+    datatype = dtype_kind_map[data.data.dtype.kind]
+
+    latlon_primary = 'Y' if 'admin0' not in data.coords else ''
+    
+    #create annotation
+    column_names=[
+        'field_name',   'group',        'display_name', 'description',      'data_type',    'units',    'units_description',    'primary',  'date_format',      'gadm_level',   'resolve_to_gadm',  'coord_format', 'qualifies',    'qualifier_role']
+    
+    rows = [
```
ccjoel, line 244: This is a very interesting diff :-)

tasks/tasks/data_modeling.py
```diff
@@ -0,0 +1,398 @@
...
+    dataset.to_netcdf('data.nc')
+    with open('data.nc', 'rb') as f:
+        data_bytes = f.read()
+    os.remove('data.nc')
+
+    
+    #make post request
+    # e.g. as a curl command:
+    # curl -v -F "metadata=@metadata.json" -F "data=@data.csv" -F "dictionary=@dictionary.csv" http://localhost:8000/indicators/register
+    response = requests.post(f'{settings.DOJO_URL}/indicators/register', files={
```
ccjoel, line 273: I'm glad this worked!

ui/client/dagpipes/overview.css
adamgilg: Yeah, I stripped back the rest of the css/sass as most of it was setting global styles on basic html tags (like `<li>` and whatnot), which then caused global styling issues. Plus I wanted to be consistent with the rest of the app. Though I have my doubts about tss-react as a permanent solution to our styling woes, but for now we're locked in.

As for these styles, they're styling the nodes generated by react-flow that we don't have direct access to. There's probably a way to directly apply styles to classes with tss-react, but it isn't worth spending the time figuring out how to do it right now (and maybe not ever... the tss-react docs are not good).

api/src/indicators.py
adamgilg: Ah, good point. Might be worth changing it to an `and` version in case a file has just one (though any file that goes through the data annotation pipeline _should_ have both, as they get set when we do the final upload steps).

ui/client/dagpipes/overview.css
ccjoel: Ah, I see. This is good context!

PR #192, opened 2023-11-07
Rename Data Modeling Group By Node to Reduce By

adamgilg:
With a light cleanup of some tech debt remaining from the switchover from the Dagpipes app -> Dojo, as well as a couple unrelated Lint fixes.

### DIFF
--- a/ui/client/dagpipes/constants.js
+++ b/ui/client/dagpipes/constants.js
@@ -14,22 +14,6 @@ export const bottomHandle = {
   backgroundColor: '#778899',
 };
 
-export const dojoMockDatasources = [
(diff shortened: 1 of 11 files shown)


ui/client/dagpipes/constants.js
```diff
@@ -14,22 +14,6 @@ export const bottomHandle = {
   backgroundColor: '#778899',
 };
 
-export const dojoMockDatasources = [
```
adamgilg, line 17: All of the chunks of code removed from constants.js weren't being used anywhere

ccjoel (approved):
+1!

PR #194, opened 2023-11-07
AI Assistant Improvements

adamgilg:
Adds a number of small improvements to the AI Assistant, including:

- An empty card when the page first loads instructing users to put their question in the box below
- A loading spinner while responses come back
- An error card with a 'try again' button
- Shortens long document titles that wouldn't fit into the provided space (and puts the full titles in a tooltip)


ui/client/components/aiAssistant/AIAssistant.js
```diff
@@ -71,67 +79,96 @@ const AIAssistant = () => {
...
-
-        setPreviousSearches((oldPairs) => ([...oldPairs, queryDetails]));
-        setSearchPhrase('');
-      } catch (error) {
-        console.error(error);
-        // Handle errors
-      }
+      // store the user input query and the responses keyed under a timestamp
+      // so we can match them up
+      const queryKey = Date.now();
```
adamgilg, line 150: Queries and responses are now broken apart (and matched up by the queryKey) so that we can save the query and display it right away regardless of whether we get a response. `performSearch` is broken out into its own function above so that we can try fetching the response again (and fill out just the response object under a given `queryKey`) if we get an error and the user clicks the "Try Again" button to retry their query.

ui/client/components/aiAssistant/AIAssistant.js
```diff
@@ -71,67 +79,96 @@ const AIAssistant = () => {
...
+      setPreviousQueries((oldQueries) => ({ ...oldQueries, [queryKey]: searchPhrase }));
+      setSearchPhrase('');
+      setResponses((oldResponses) => ({
+        ...oldResponses,
+        [queryKey]: {
+          data: null,
+          status: 'loading',
+        }
+      }));
+      setAnyResponseLoading(true);
```
adamgilg, line 160: We still have a global anyResponseLoading state because we use this to determine whether certain things are disabled/displayed at the component level, not just within the map over queries/responses.

ccjoel (approved):
Deploy and sell it already!

ccjoel:
Also- shut up and take my money!

PR #196, opened 2023-11-08
Data Transformation: Make Geo & Temporal Resolution Not Required

adamgilg:
It turned out this was slightly more complicated than just removing the `required` props and the disabled checks, as the DataTransformation component was relying on the RunTransformation component always running. However, this didn't take much tweaking to fix. 

I also fixed the transformation processing overlay styling.

### DIFF
--- a/ui/client/datasets/DataTransformation/DataTransformation.js
+++ b/ui/client/datasets/DataTransformation/DataTransformation.js
@@ -548,16 +548,6 @@ const DataTransformation = ({
       return true;
     }
 
-    if (!mapResolutionError && !savedMapResolution) {
-      // disable if we are requiring a map resolution to be chosen and it hasn't been
-      return true;
-    }
-
-    if (!timeResolutionError && !savedTimeResolution) {
-      // disable if we are requiring a time resolution to be set and it hasn't been
-      return true;
-    }
-
     if (!gadmResolutionError && !savedGADMOverrides) {
       return true;
     }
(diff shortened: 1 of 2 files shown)


ui/client/datasets/DataTransformation/DataTransformation.js
```diff
@@ -664,7 +652,10 @@ const DataTransformation = ({
   };
 
   const processGadmOverrides = () => {
-    transformationsRef.current.overrides = { gadm: savedGADMOverrides };
+    // only add GADM overrides to the transformationsRef if they are present
```
adamgilg, line 655: We use this ref to determine whether we want to update the annotations object, so if we add this here then we end up updating the annotations object with `overrides: { gadm: null }`, which seems pointless.

ui/client/datasets/DataTransformation/RunTransformations.js
```diff
@@ -40,7 +40,7 @@ const RunTransformations = ({
   setAllTransformationsComplete,
   setTransformationProcessingError,
 }) => {
-  const classes = useStyles();
+  const { classes } = useStyles();
```
adamgilg, line 43: These styles weren't getting applied at all previously because of the incorrect import. Now they are and it looks better.

ui/client/datasets/DataTransformation/DataTransformation.js
```diff
@@ -692,6 +683,11 @@ const DataTransformation = ({
     setJobsConfig(transformations);
     // The actual jump to next step happens in the useEffect at the top of this function
     // after the jobs have completed in RunTransformation
+
+    // if there are no transformations, toggle this to trigger our 'handleNext' useEffect at top
```
adamgilg, line 687: We need to do this now as we are no longer forcing the user to complete at least one transformation, so with no transformations we'll never get inside the `if` block in the handleNext useEffect at the top.

ccjoel (commented):
👍🏼

PR #201, opened 2023-11-16
GET job status no side effects, Composite Dataset Register endpoint improved

ccjoel:
* Now API-first composite dataset register endpoint (`C.DR.E`?) returns the url to check status of the dataset register process finalizing.

* The rqworker tasks that finishes registration now returns the dataset with the published property set to true (the final dataset used to have published=true, but not on the job output from rqworker, which was confusing.

* New GET endpoints to get job result, job status; before POST both created and fetched status/result, which is dangerous (could create new jobs by mistake, can cause side-effects). 

* Some misc lint fixes, and added comments/docstrings to some fn and handlers.

* Added recent job entries to `available_job_strings` endpoint. This should help Dojo maintainers when wanting to quickly list available job fully-qualified names without having to resort to memory or looking at `tasks/tasks/*.py` source files.


api/src/data.py
```diff
@@ -149,9 +201,70 @@ def cancel_job(job_id):
...
+        "started_at": job.started_at,
+        "status": job.get_status(),
+        "job_error": job.exc_info,
+        "result": job.result,
+    }
+    return response
+
+
+# TODO Use instead of Optional Dict|None (which doesn't generate swagger)
+class JobCreateOptions(BaseModel):
```
ccjoel, line 249: Did not muster the courage to mess with the create `job` fn below, but this is a start- we can try to add it so that swagger displays better documentation/sample, and for developer happiness, of course.

PR #205, opened 2023-11-20
Able to download raw file and standardized files

ccjoel:


ui/client/components/CSVDownload.js
```diff
@@ -18,44 +19,59 @@ import { makeStyles } from 'tss-react/mui';
 import BasicAlert from './BasicAlert';
 
 const useStyles = makeStyles()((theme) => ({
-  root: {
-    height: '114px',
-  },
+  root: { },
```
ccjoel, line 22: Removing the fixed height addressed the main responsive issue (Wasnt rendering nicely on smaller screens).

PR #208, opened 2023-12-06
Dojo Docs in Docusaurus

adamgilg:
This adds a copy of the [Dojo Docs](https://github.com/dojo-modeling/dojo-docs/) into this repo. Now they're rebuilt with Docusaurus, with a general style overhaul and more to come. 

We'll have to set up Github Pages deployment to dojo-modeling.com from the `gh-pages` branch (we could also use `main`, but `gh-pages` or otherwise is suggested as an alternative for keeping Pages deployments separate from every `main` merge). 

The only thing this implementation is missing from the previous docs, at the moment, is the search bar. This still needs to be implemented and will take a little more work. 

Other than that, this PR is ready to go. 

I also added this to the intro page:

Screenshot 2023-12-06 at 5 55 10 PM

Let me know if you want to change the wording/etc.

There is also no Docker setup yet for running this locally, as it's small and simple and doesn't interact with anything else. We can add that if we feel the need for it sooner or down the road/never if we don't.


dojo-docs/README.md
```diff
@@ -0,0 +1,41 @@
...
+### Installation
+
+```
+$ npm install
+```
+
+### Local Development
+
+```
+$ npm run start
```
ccjoel, line 14: So easy!

dojo-docs/docs/FAQ.md
```diff
@@ -0,0 +1,40 @@
...
+layout: default
+title: Frequently Asked Questions
+---
+
+# Dojo / World Modelers FAQ
+
+## 1. Will my model be a good fit for running in Dojo?
+To be a good fit for Dojo your model should be able to run in a Debian/Ubuntu based environment.
+
+It is also useful, but not necessary, to make the most relevant input parameters perturbable before the model run which allows analysts to run “what-if” scenarios with your model.
```
ccjoel, line 11: This is a note for after this PR (preexisting "issue"):

There's this article (when running docusaurus locally):

`http://localhost:3001/details/data-format`

-It explains the format that the dataset needs to be in in order to register it. It contains details such as the requirement to possibly manually transpose the dataset data in order for Dojo to process (if the columns/rows are not in what we expect).

I think it might be good to point out that a model output should also follow this convention (Verify?). Mostly pouring my reminder here. We can discuss/assess after merging this PR. Might be useful to point our for model output, if applicable. Maybe not in the FAQ, or maybe a quick link from this FAQ to it (for the model output shape requirement) since it doesn't hurt to have this redundancy.

ccjoel:
I noticed that, when expanding or opening sub articles in the left tree view, the navigation is always `/details/sub-tree-title` regardless of where in the navigation the article is at:

Screenshot 2023-12-07 at 9 55 56 AM

As an example on the above image, I would expect the url to be
http://localhost:3001/model-registration/jupyter-notebooks  (or just `jupyter`, but under `/model-registration`).

Update while still writing this comment:
I just checked that dojo-modeling.com behaves this way as well, so we may also address this in the future if we think it's worth doing (changes urls).

ccjoel:
Some extremely minor improvements we could address later:

- When scrolling on long articles, the left navigation scrolls as well. I don't see a need for this- as a UX improvement (nice-to-have, after this PR) we could scroll the middle content only and allow the left/right menu to remain static- so as to have all the articles remain into view).
- Similarly related to scrolling: while scrolling the middle contents, the right-side article table of contents auto-highlights (in green) the title we scrolled past. I can also click on each article on this right pane to scroll to the contents. However, when we get to the end of the page the headings will not highlight the last item, and clicking on the items at the end might not highlight at all (since we got to the end of the page and it is bound to scroll % looks like). The fact that we can never click/make those green is weird, but is an extremely minor detail.

Leaving a recording:

https://github.com/jataware/dojo/assets/1967061/a150a6cf-f310-4d4d-97c8-8291eabdf956

Note: I wonder if both are a limitation to docusaurus itself. If so, maybe we won't address them at all. Just going trigger-happy pointing minor improvements!

dojo-docs/docs/dojo-api.md
```diff
@@ -0,0 +1,125 @@
...
+
+This document outlines how to interact with the Dojo API to fetch models, execute models, and fetch model runs.
+
+## Model Discovery
+
+You can search Dojo for models using the `GET /models`. For example, to find models related to `crops` you would send:
+
+```
+curl -X 'GET' \
+  'https://api.dojo-modeling.com/models?query=crops&size=10' \
```
ccjoel, line 17: Another reminder- possible pre-existing issue:

I think this url is old/wrong. Likely existing issue on dojo-modeling.com. I wonder if it is just an example and not meant to be tried out, but maybe we need a pro-tip kind-of note stating that all these urls are examples for when others deploy/self-host the app?

Another option is to update these to the correct urls where they are currently deployed at. It would have auth around them, but our existing transition users could copy them and use the endpoints seamlessly from the docs. Thoughts?

dojo-docs/docs/model-execution.md
```diff
@@ -0,0 +1,428 @@
...
+```
+
+## CLI help
+
+The following commands will provide details of each available dojo command:
+
+```
+dojo --help
+dojo describe --help
+dojo listmodels --help
```
ccjoel, line 85: Another pre-existing problem I'd like to note here, while inspired, is that I'm not sure how to really run these `dojo` commands. They may be out of date. For example, they are not available when running under the terminal emulator on the Dojo UI:

Screenshot 2023-12-07 at 10 17 06 AM

Installing the `dojo` cli locally fails most of the time (I tried it once), and it is not clear to me if it properly integrates with a running Dojo API running on the same host, nor if it works, nor if we have tested this since a couple of years back.

Possibly a discussion point- should we hide these dojo-cli documentation? The less not-as-useful documentation we add, the less confusing this complex system might be to new eyes.

Like the rest of these comments, we can discuss after deploying the new docs.

ccjoel: cc @brandomr in case you see this mention and would like to add some color to this.

ccjoel (approved):
+1, amazing work. All comments are left are related to pre-existing documentation or picky improvements we should leave for later.

adamgilg:
> Some extremely minor improvements we could address later:
> 
>     * When scrolling on long articles, the left navigation scrolls as well. I don't see a need for this- as a UX improvement (nice-to-have, after this PR) we could scroll the middle content only and allow the left/right menu to remain static- so as to have all the articles remain into view).
> 
>     * Similarly related to scrolling: while scrolling the middle contents, the right-side article table of contents auto-highlights (in green) the title we scrolled past. I can also click on each article on this right pane to scroll to the contents. However, when we get to the end of the page the headings will not highlight the last item, and clicking on the items at the end might not highlight at all (since we got to the end of the page and it is bound to scroll % looks like). The fact that we can never click/make those green is weird, but is an extremely minor detail.
> 
> 
> 
> Note: I wonder if both are a limitation to docusaurus itself. If so, maybe we won't address them at all. Just going trigger-happy pointing minor improvements!

These are definitely both parts of Docusaurus. 

1. It's possible there's a flag specifically to turn your first point off (left nav scrolling), but I don't particularly see it as a problem. If we disable that, then we're left with a big blank space over there, and if someone wants to navigate elsewhere, they're then forced to scroll back to the top of the page to do so. 

2. I believe what's happening with your second point (without looking at any of the code for the right TOC highlights) is that it's only able to highlight whatever is at the top of the viewport. If you look at the FAQ page, it will only highlight the number that is fully visible at the top of the screen. If you zoom in or out, it will change what is highlighted based on what's fully visible. Clicking on the TOC links still changes the URL query params. I would lean towards not messing around with this, as I think it would be quite a bit more work than it's worth to attempt to 'fix' it, whatever that would even look like.

dojo-docs/docs/model-execution.md
adamgilg: 100% agree. We need an overhaul of a lot of the terminal and model docs.

adamgilg:
> I noticed that, when expanding or opening sub articles in the left tree view, the navigation is always `/details/sub-tree-title` regardless of where in the navigation the article is at:
> 
> As an example on the above image, I would expect the url to be http://localhost:3001/model-registration/jupyter-notebooks (or just `jupyter`, but under `/model-registration`).
> 
> Update while still writing this comment: I just checked that dojo-modeling.com behaves this way as well, so we may also address this in the future if we think it's worth doing (changes urls).

Yeah, I moved all the sub-pages into a /details folder for organizational purposes recently in the old docs. This means that due to the simple url path structure they're all /details/etc. We can probably manually change this, but I haven't looked into how to set that up yet.

PR #224, opened 2023-12-18
Navigation Redesign

adamgilg:
This PR has a site wide navigation redesign feature: the top navbar and collapsible sidebar. It also features a slight main landing page refresh, as well as new sub-landing pages for Models, Datasets, and Documents. 

Screenshot 2023-12-18 at 4 52 35 PM

Screenshot 2023-12-18 at 4 52 49 PM
Screenshot 2023-12-18 at 5 02 27 PM

Screenshot 2023-12-18 at 5 04 16 PM


brandomr:
@adamgilg this looks great as a first pass! a couple thoughts--

* i think we `an open source platform...` to just `a platform` -- not that it _isn't_ open source -- just that I worry some non-technical users will think that everything they do on the platform will be _public_ and that might trigger some alarm bells.
* in the documents section i think we should nix the term `PDF` and just keep it very generic since we were asked to add support for `ppt` and `docx` so would need to update that language anyway
* for model runs I think we should add a 2nd sentence: `Retrieve model execution or simulation run results for further analysis`
* for dataset search under `view existing datasets` we should mention (perhaps in an additional 2nd sentence?) that `users may search for datasets based on columns/feature names or the dataset's description`

PR #230, opened 2024-01-02
Geotiff support in data modeling

david-andrew:
- Added support for geotiff data in data modeler
- created debug helper functions for plotting data from within the docker container


ccjoel (approved):
🪨  (its a rock; after saving it gave me crumbled paper vibes and that didn't look good!)

PR #233, opened 2024-01-03
Caddy server setup

ccjoel:
The purpose of this work is to add a built-in prod-like router within the dojo-stack. Currently for staging/prod/deployments we use an external nginx server with a mix of in-code and in-flight configuration. Setting this up would mean we can start the stack without the need to wrap with nginx or other servers separate from our stack.

Uses a `prodv2` docker profile. We can/could merge, try it out, and either eventually remove the profile or extend our docker profiles usage.

All of it will run by running the following:
```
docker compose --profile prodv2 up -d
```


router/Caddyfile
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    redir /api/dojo /api/dojo/
+    redir /openapi.json /api/dojo/openapi.json
+    handle_path /api/dojo/* {
+        reverse_proxy {$DOJO_HOST}:80
+    }
+
+    redir /api/ai-docs /api/ai-docs/
```
ccjoel: We can remove this redirect rule, as boxytas/knowledge is part of the Dojo API now (which is covered by the above `/api/dojo` rule).

docker-compose.build-override.yaml
```diff
@@ -218,7 +220,38 @@ services:
       - dojo-stack
     ports:
       - 8012:8012
+    # profiles:
```
ccjoel, line 223: Suggestion, where we could enable testmos on local dev only.

docker-compose.build-override.yaml
```diff
@@ -193,6 +193,8 @@ services:
       - ${PWD}/data:/data
     networks:
       dojo-stack: {}
+    # profiles:
```
ccjoel, line 196: Suggestion, where we could enable minio on local dev only.

PR #260, opened 2024-02-27
New Data Modeling Nodes

david-andrew:
### Added backend support for new data modeling nodes/features
- `Threshold` node:
    - optionally preserving NaNs from data (generates float mask instead of bool mask)
    - can accept percentiles in addition to raw values for threshold, via `xx%` syntax
- `Join` node
    - dropdown for selecting what type of operation to use when joining (add/subtract/multiply/divide/power). 
        > Note that non-commutative operators (subtract, divide, power) need to be attached left to right since currently there's no way to tell which parent is attached to the left node, and which is attached to the right node
- `Reduce By` node
    - dropdown to select aggregation operation (sum, mean, median, min, max, mode, std)
- **new** `Scalar Operation` node
- **new** `Select Slice` node
- **new** `Mask to Distance Field` node

### Other stuff
- update flowcast library version to include new features
- some minor UI changes to support new node features
    - added `Preserve NaN` checkbox to `threshold` node
    - made text field for `threshold` node take text instead of number to support percentiles as input
    - added `subtract` option to `scalar operation` node
- better error printout with traceback to log when data modeler job fails

### DIFF
--- a/tasks/tasks/data_modeling.py
+++ b/tasks/tasks/data_modeling.py
@@ -415,11 +415,12 @@ def unhandled_run_flowcast_job(context:FlowcastContext) -> dict:
             value = node['data']['input']['value']
             threshold_type = ThresholdType[node['data']['input']['type']]
             if '%' in value:
-                raise NotImplementedError('Percentile thresholds are not yet supported')
-                value = float(value.strip('%')) / 100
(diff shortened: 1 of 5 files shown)


adamgilg (approved):
Looks good!

PR #262, opened 2024-02-29
UI/Data Modeling: Capture Left/Right Handle in Join Node Data

adamgilg:
This PR updates the data we pass to the server to include the `target_handle` if it's part of an edge's data. This is only relevant to the Join (multiply) node, which has `multiply-handle-1/2` for left/right. 

I also updated React Flow to the latest version and swapped out a deprecated function for the new version with no functionality change.

### DIFF
--- a/ui/client/dagpipes/PipeEditor.js
+++ b/ui/client/dagpipes/PipeEditor.js
@@ -279,7 +279,8 @@ const PipeEditor = () => {
       }
 
       const reactFlowBounds = reactFlowWrapper.current.getBoundingClientRect();
-      const position = reactFlowInstance.project({
+      // subtracting the bounds is not required, but it positions the drag buttons better on drop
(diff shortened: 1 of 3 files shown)


ui/client/dagpipes/PipeEditor.js
```diff
@@ -278,10 +278,9 @@ const PipeEditor = () => {
         return;
       }
 
-      const reactFlowBounds = reactFlowWrapper.current.getBoundingClientRect();
-      const position = reactFlowInstance.project({
-        x: event.clientX - reactFlowBounds.left,
-        y: event.clientY - reactFlowBounds.top,
+      const position = reactFlowInstance.screenToFlowPosition({
```
adamgilg, line 281: https://reactflow.dev/api-reference/types/react-flow-instance#project

PR #278, opened 2024-08-20
Track user chat sessions in Redis

brandomr:
# Main Changes

The goal of this PR is twofold:
1. If a user navigates off the page and returns, their is some history of their chat
2. Two users chat contexts will no longer intermingle. `User B` cannot ask: "What did I last ask you?" to see `User A`'s query.

This is accomplished by:

1. Adding a `chatuserToken` to the user's `localstorage`
5. Pass in that token to `/knowledge/chat` as a query parameter
6. Fetch the message history from Redis for that `chatuserToken`
7. Instantiate a new `librarian` with those `messages` as history
8. Respond to the user's query
9. Overwrite any stored messages for that `chatuserToken` in Redis with a 24 hour TTL

This seems to work well enough but I am concerned about a couple things:

1. Is using `localstorage` the right way to do this? In production I assume we'd want to tie things to the username--so this feels a bit hacky. But I don't think we need real persistence of these chats.
2. If the chats get very long in a given session this could blow up Redis. `messages` might need to be truncated somehow.

@ccjoel I'm curious your take on 1) and @david-andrew I'm curious your thoughts on 2).

## Secondary Changes

This PR also adds a fuzzy and wildcard search over titles. This enables the agent to do things like take a user query:

```
Can you compare the World Situation Report's findings with that of the UN contributions under paris agreement?
```

and correctly retrieve two documents:

1. `World Economic Situation and Prospects 2024` 
2. `Nationally determined contributions under the Paris Agreement`

This may be sufficient without needing to implement further semantic search over titles.

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -46,4 +46,4 @@ idna==2.10
 sniffio==1.3.0
 httpcore==1.0.5
 httpx==0.27.0
-jatarag==0.1.5
+jatarag==0.1.8
--- a/api/src/knowledge.py
+++ b/api/src/knowledge.py
@@ -8,6 +8,7 @@ from elasticsearch import Elasticsearch
 from fastapi import (
     APIRouter,
     HTTPException,
+    Header,
     # Response,
     status,
     # UploadFile,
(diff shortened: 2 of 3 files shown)


david-andrew:
yeah I think 2) is fine. making a new instance of the librarian for every call should be pretty lightweight, so passing in the current messages state each API call should be good

PR #280, opened 2024-08-28
fix data modeling errors not being logged

david-andrew:
Fixes data modeling not logging exceptions that occur during execution

for some reason `traceback.print_exc()` no longer prints to the container logs, so switching to a regular print statement so that errors are logged properly


brandomr (approved):
tested this out, do see better logs now and nothing seems broken

